PR #8, opened 2022-04-10
Solution

vldtsw:


taxi/urls.py
```diff
@@ -1,9 +1,26 @@
...
+         name="car-list"),
+    path("cars/<int:pk>",
+         CarDetailView.as_view(),
+         name="car_detail"),
+    path("drivers/",
+         DriverListView.as_view(),
+         name="driver-list"),
+    path("drivers/<int:pk>",
+         DriverDetailView.as_view(),
+         name="driver_detail")
```
liuda-chuzdiuk, line 23: Could you please use one seperator in path name ("_")

liuda-chuzdiuk (changes requested):

taxi/urls.py
vldtsw: Don't really understand. Can you explain a bit more detailed?

liuda-chuzdiuk (approved):

IvanRamyk (approved):
Good job!

PR #11, opened 2022-04-13
'Solution'

8Men5Camels:


templates/includes/pagination.html
```diff
@@ -0,0 +1,21 @@
...
+        </li>
+
+        {% if page_obj.has_next %}
+            <li class="page-item">
+                <a href="?page={{ page_obj.next_page_number }}" class="page-link">next</a>
+            </li>
+        {% endif %}
+
+    </ul>
+{% endif %}
```
liuda-chuzdiuk, line 21: please add new line in the end and check other files

liuda-chuzdiuk (changes requested):

templates/includes/pagination.html
8Men5Camels: done

IvanRamyk (approved):
Well done!

PR #20, opened 2022-06-30
Solution

YuriiKindrat:


taxi/views.py
```diff
@@ -17,3 +18,31 @@ def index(request):
     }
 
     return render(request, "taxi/index.html", context=context)
+
+
+class ManufacturerListView(generic.ListView):
+    model = Manufacturer
+    template_name = "taxi/manufacturer_list.html"
+    context_object_name = "manufacturer_list"
+    queryset = Manufacturer.objects.all()
```
Flashmobber, line 27: no need to create a queryset for this view

taxi/views.py
```diff
@@ -17,3 +18,31 @@ def index(request):
...
+    paginate_by = 2
+
+
+class CarListView(generic.ListView):
+    model = Car
+    paginate_by = 2
+    queryset = Car.objects.all().select_related("manufacturer")
```
Flashmobber, line 34: you can use select_related directly to objects

taxi/views.py
```diff
@@ -17,3 +18,31 @@ def index(request):
...
+
+
+class DriverListView(generic.ListView):
+    model = Driver
+    queryset = Driver.objects.all()
+
+
+class DriverDetailView(generic.DetailView):
+    model = Driver
+    queryset = Driver.objects.all().prefetch_related("cars__manufacturer")
```
Flashmobber, line 48: the same with prefetch_related, no need to use all()

taxi/views.py
Flashmobber: add pagination to all ListView

Flashmobber (changes requested):

Flashmobber (approved):

MasterpieceElbow (approved):

PR #23, opened 2022-07-02
All Done.

Flashmobber:


Flashmobber:
image
image
image
image
image
image

taxi/views.py
```diff
@@ -1,8 +1,40 @@
...
-
+from django.views import generic
 from .models import Driver, Car, Manufacturer
 
 
+class ManufacturerListView(generic.ListView):
+    model = Manufacturer
+    template_name = "taxi/manufacturers.html"
+    context_object_name = "brands"
+    paginate_by = 10
```
Magus6784, line 10: In task we can see:
Set pagination equals to 2.
But if you have more tests info i think no one care

taxi/views.py
```diff
@@ -1,8 +1,40 @@
...
+
+
+class CarListView(generic.ListView):
+    model = Car
+    template_name = "taxi/cars_list.html"
+    context_object_name = "cars"
+    queryset = Car.objects.select_related("manufacturer")
```
Magus6784, line 17: image
Without 1000 words
This also true for files name like "drivers_list", "cars_list"

templates/taxi/car_detail.html
```diff
@@ -0,0 +1,20 @@
...
+    <hr>
+    <p>Car drivers:</p>
+    {% for driver in car.drivers.all %}
+        <li>
+            {% if driver.first_name %}
+                <a href="{{ driver.get_absolute_url }}">{{ driver.first_name }} {{ driver.last_name }}</a>
+                {% else %}{{ driver.username }}
+            {% endif %}
+        </li>
+    {% endfor %}
```
Magus6784, line 19: Better add "empty" situation message

templates/taxi/car_detail.html
```diff
@@ -0,0 +1,20 @@
+{% extends "base.html" %}
+
+{% block title %}
+    <title>car: {{ car.manufacturer.name }} {{ car.model }}</title>
+{% endblock %}
+
+{% block content %}
+    <h1>{{ car.manufacturer.name }} {{ car.model }}</h1>
+    <p>country: {{ car.manufacturer.country }}</p>
+    <hr>
```
Magus6784, line 10: This <hr> must be in {% for %} because with this line we split up drivers

Magus6784: Ouh
hr, i mean hr

templates/taxi/driver_detail.html
```diff
@@ -0,0 +1,24 @@
...
+
+{% block title %}
+    <title>Driver: {{ driver.username }}</title>
+{% endblock %}
+
+{% block content %}
+    <h1>
+        {% if driver.first_name %} {{ driver.first_name }} {{ driver.last_name }}
+            {% else %}{{ driver.username }}
+        {% endif %}
```
Magus6784, line 11: I don't sure about it
We can have first_name, but don't have last_name

templates/taxi/driver_detail.html
```diff
@@ -0,0 +1,24 @@
...
+    </h1>
+    <div>
+        <div>username: {{ driver.username }}</div>
+        {% if driver.licence_number %}
+            <div>licence: {{ driver.licence_number }}</div>
+        {% endif %}
```
Magus6784, line 17: We alwais have licence_number, it's can be null or blank

templates/taxi/driver_detail.html
```diff
@@ -0,0 +1,24 @@
...
+    </div>
+    <hr>
```
Magus6784, line 19: Same for this hr

templates/taxi/car_detail.html
Magus6784: image

Magus6784 (changes requested):

taxi/views.py
Flashmobber: I know that it's 2 in the task, but I have got more data to view and the point is not just to follow the instruction but to show you understand how to work with this parameter.

templates/taxi/car_detail.html
Flashmobber: I did it intentionally because it looks better to me. It doesn't mean I don't understand the task.

templates/taxi/car_detail.html
Flashmobber: Built-in-tag 'empty' can be used along with 'for' only. there is no such an option with 'if'.

templates/taxi/driver_detail.html
Flashmobber: Agreed. But if 'last name' remains empty we'll get an empty value. As a result, only 'first_name' will be viewed.

templates/taxi/driver_detail.html
Flashmobber: image
Absolutely can as you see.

templates/taxi/driver_detail.html
Flashmobber: Same for that.

taxi/views.py
Flashmobber: If the meaning is 'a list of cars' the better option is to use the plural like I did.

templates/taxi/driver_detail.html
Magus6784: Lol

Magus6784 (approved):

MasterpieceElbow (approved):

PR #27, opened 2022-07-04
Solution

Hectorovich:


templates/taxi/car_list.html
```diff
@@ -0,0 +1,31 @@
+{% extends "base.html" %}
+
+{% block title %}<title>Cars</title>{% endblock %}
+
+{% block content %}
+    <h2>List of cars</h2>
+
+    {% if car_list %}
+        <style>
```
katevasilenko, line 9: maybe it's better to add to static/css, of course, if it's possible

katevasilenko (changes requested):
You did great, but add ordering to models

katevasilenko (approved):

MasterpieceElbow (approved):

PR #36, opened 2022-07-04
Solution

Chardona:
image
image
image
image
image
image

### DIFF
--- a/taxi/urls.py
+++ b/taxi/urls.py
@@ -1,9 +1,30 @@
 from django.urls import path
 
-from .views import index
+from .views import index, ManufacturerListView, DriverListView, CarListView, CarDetailView, DriverDetailView
 
 urlpatterns = [
     path("", index, name="index"),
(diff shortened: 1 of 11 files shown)


darten360 (approved):

taxi/urls.py
```diff
@@ -1,9 +1,30 @@
 from django.urls import path
 
-from .views import index
+from .views import index, ManufacturerListView, DriverListView, CarListView, CarDetailView, DriverDetailView
 
 urlpatterns = [
     path("", index, name="index"),
+    path("drivers/",
+         DriverListView.as_view(),
+         name="driver-list-view"
```
MasterpieceElbow, line 9: 1. "-view" in names is redundant because you repeat it everywhere.
2. It is actually not a view but a path.

MasterpieceElbow (changes requested):

TPolina (approved):

PR #63, opened 2022-08-11
tests added first commit

DmytroSvirsa:


tests/test_taxi_view.py
```diff
@@ -0,0 +1,173 @@
...
+
+        response = self.client.get(
+            CAR_LIST_URL,
+            {"page": 2}
+        )
+
+        self.assertEqual(
+            response.status_code,
+            200
+        )
```
danylott, line 121: I believe this test does not check anything.
Because this code will work even without pagination added

tests/test_taxi_view.py
```diff
@@ -0,0 +1,173 @@
...
+        self.assertEqual(
+            list(response.context["manufacturer_list"]),
+            [self.manufacturer1]
+        )
```
danylott, line 83: Where are you checking, that it is sorted by name?

tests/test_taxi_view.py
```diff
@@ -0,0 +1,173 @@
...
+
+    def test_driver_list_has_pagination(self):
+        self.driver3 = get_user_model().objects.create_user(
+            username="dustin.henderson",
+            license_number="DUS25131",
+            first_name="Dustin",
+            last_name="Henderson",
+            password="3edcbgt5"
+        )
```
Nattalli, line 155: is it necessary to add a new page?

danylott (changes requested):

tests/test_taxi_view.py
```diff
@@ -0,0 +1,173 @@
...
+        self.car1 = Car.objects.create(
+            model="test_name_1",
+            manufacturer=self.manufacturer1
+        )
+        self.car2 = Car.objects.create(
+            model="Cadillac Escalade",
+            manufacturer=self.manufacturer2
+        )
+        self.car1.drivers.add(self.driver1)
+        self.car2.drivers.add(self.driver2)
```
danylott, line 47: You can use here your fixtures:
https://stackoverflow.com/a/2937349

And also you should split this 1 test class to 3 separate:
for Car, for Manufacturer & for Driver

PR #88, opened 2022-10-20
Solution

Vlad-Peresta:


Vlad-Peresta:
Screenshots of all created or modified pages:
image
image
image
image
image

templates/includes/pagination.html
```diff
@@ -0,0 +1,25 @@
...
+    {% if page_obj.has_next %}
+      <li class="page-item">
+        <a href="?page={{ page_obj.next_page_number }}"
+           class="page-link">
+           next
+        </a>
+      </li>
+    {% endif %}
+  </ul>
+{% endif %}
```
DmytroSvirsa, line 25: let's add info about how many pages we have at all

DmytroSvirsa (changes requested):

templates/includes/pagination.html
Vlad-Peresta: I agree, it would be more comfortable

Vlad-Peresta: image

DmytroSvirsa (approved):

PR #97, opened 2022-10-24
Solution

wQuelS:


IvanStored (approved):

taxi/urls.py
```diff
@@ -1,9 +1,39 @@
...
+        ManufacturerListView.as_view(),
+        name="manufacturer-list"),
+    path(
+        "cars/",
+        CarListView.as_view(),
+        name="car-list"),
+    path(
+        "cars/<int:pk>",
+        CarDetailView.as_view(),
+        name="car-detail",
```
Nattalli, line 25: does this URL work fine? I think that it's better to close the path with "/" in the end:
`path("cars/<int:pk>/")`
WDYT?

taxi/views.py
```diff
@@ -17,3 +18,31 @@ def index(request):
...
+    model = Manufacturer
+    template_name = "taxi/manufacturer_list.html"
+    context_object_name = "manufacturer_list"
+    queryset = Manufacturer.objects.order_by("name")
+    paginate_by = 5
+
+
+class CarListView(generic.ListView):
+    model = Car
+    queryset = Car.objects.all().select_related("manufacturer")
```
Nattalli, line 33: remove `.all'()`

taxi/views.py
```diff
@@ -17,3 +18,31 @@ def index(request):
...
+
+
+class DriverListView(generic.ListView):
+    model = Driver
+    paginate_by = 5
+
+
+class DriverDetailView(generic.DetailView):
+    model = Driver
+    queryset = Driver.objects.all().prefetch_related("cars__manufacturer")
```
Nattalli, line 48: remove `.all()`

taxi_service/settings.py
```diff
@@ -37,11 +41,13 @@
     "django.contrib.sessions",
     "django.contrib.messages",
     "django.contrib.staticfiles",
+    "debug_toolbar",
```
Nattalli, line 44: wow, do you use it?

templates/taxi/car_list.html
```diff
@@ -0,0 +1,16 @@
+{%  extends "base.html" %}
+
+{% block content %}
+  <h1>Car List</h1>
+  {% if car_list %}
+    <ul>
+      {% for car in car_list %}
+        <li class="list-stuff">
+          <a href="{% url 'taxi:car-detail' pk=car.pk %}">{{ car.model }}</a>
```
Nattalli, line 9: I understand that it would have a less beautiful interface, but you had to do it this way:
`In the car list set a link on the id field that leads to the car detail page.`

Nattalli (changes requested):

Nattalli:
Left some small comments

taxi_service/settings.py
wQuelS: I did

Nattalli (approved):

PR #121, opened 2022-11-02
Solution

KatrinLazarenko:
image
image
image
image
image

### DIFF
--- a/taxi/views.py
+++ b/taxi/views.py
@@ -1,4 +1,6 @@
+from django.http import Http404
 from django.shortcuts import render
+from django.views import generic
 
 from .models import Driver, Car, Manufacturer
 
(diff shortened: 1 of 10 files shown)


shurushku (approved):
Looks good

taxi/views.py
```diff
@@ -17,3 +19,31 @@ def index(request):
...
+    paginate_by = 5
+
+
+class CarDetailView(generic.DetailView):
+    model = Car
+
+
+class DriverListView(generic.ListView):
+    model = Driver
+    queryset = Driver.objects.all().prefetch_related("cars")
```
TPolina, line 44: you don't need queryset here. You need it in DriverDetailView
also, you need to optimize getting manufacturers of the cars as well:
https://docs.djangoproject.com/en/4.1/ref/models/querysets/#prefetch-related

templates/taxi/car_detail.html
```diff
@@ -0,0 +1,14 @@
...
+  <h1>Car: {{ car.model }}</h1>
+  <p><strong>Manufacturer: {{ car.manufacturer.name }}</strong></p>
+
+  <div class="ml-3">
+    <h3>Car`s drivers</h3>
+  {% for driver in car.drivers.all %}
+    <hr>
+    <p><strong>Driver name: {{ driver.first_name }} {{ driver.last_name }}</strong></p>
+  {% endfor %}
+  </div>
```
TPolina, line 13: it's better to add a separate phrase in case there are no drivers

templates/taxi/driver_detail.html
```diff
@@ -0,0 +1,17 @@
...
+  <p><strong>License: {{ driver.license_number }}</strong></p>
+
+  <div class="ml-3">
+    <h3>Driver`s cars</h3>
+  {% for car in driver.cars.all %}
+    <hr>
+    <p><strong>Model: {{ car.model }}</strong></p>
+    <p><strong>Manufacturer: {{ car.manufacturer.name }}</strong></p>
+  {% endfor %}
+  </div>
```
TPolina, line 16: it's better to add a separate phrase in case there are no cars

TPolina (changes requested):

TPolina (approved):
Well done!

PR #123, opened 2022-11-02
Solution

matechris:

### DIFF
--- a/taxi/views.py
+++ b/taxi/views.py
@@ -1,4 +1,5 @@
 from django.shortcuts import render
+from django.views import generic
 
 from .models import Driver, Car, Manufacturer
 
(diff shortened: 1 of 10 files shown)


templates/taxi/car_list.html
```diff
@@ -0,0 +1,12 @@
+{% extends "base.html" %}
+
+{% block content %}
+  <h1>Car list</h1>
+
+  <ul>
```
HStakhiv, line 6: What if you have no car in your service? I think it is better to add `if` and `else`. What do you think?

HStakhiv (changes requested):
Ok. But..

HStakhiv (approved):
Nice!

taxi/views.py
```diff
@@ -17,3 +18,31 @@ def index(request):
...
+
+
+class DriverListView(generic.ListView):
+    model = Driver
+    paginate_by = 5
+
+
+class DriverDetailView(generic.DetailView):
+    model = Driver
+    queryset = Driver.objects.prefetch_related("cars")
```
Y-Havryliv, line 48: did you look at the number of queries? and does this `prefetch_related` work?

templates/taxi/car_list.html
```diff
@@ -0,0 +1,16 @@
+{% extends "base.html" %}
+
+{% block content %}
+  <h1>Car list</h1>
+
+  {% if car_list %}
+    <ul>
+      {% for car in car_list %}
+          <li><strong>Manufacturer:</strong> {{ car.manufacturer.name }}</li>
+          <strong>Model:</strong> <a href="{% url 'taxi:car-detail' pk=car.id %}">{{ car.model }} </a>
```
Y-Havryliv, line 10: In the car list set a link on the id field that leads to the car detail page.

templates/taxi/car_list.html
```diff
@@ -0,0 +1,16 @@
+{% extends "base.html" %}
+
+{% block content %}
```
Y-Havryliv, line 3: `Create table with the information of each instance.`

Y-Havryliv (changes requested):

Y-Havryliv (approved):
Well done!

PR #125, opened 2022-11-02
Created class based views, pagination

alexzubkoff:


alexzubkoff:
cars
cars-int
drivers
drivers-int
manufacturers
taxi_home_page

alexzubkoff:
car_detail
cars
driver_detail
drivers
manufacturers
taxi_home_page

VadimSabadyr (approved):
for me looks good) 👍

alexzubkoff:
Thanks!

Чт, 3 нояб. 2022 г. в 13:50, VadimSabadyr ***@***.***>:

taxi/views.py
```diff
@@ -17,3 +18,39 @@ def index(request):
     }
 
     return render(request, "taxi/index.html", context=context)
+
+
+def manufacturer_list_view(request):
```
Y-Havryliv, line 23: seems like you can delete it

taxi_service/settings.py
```diff
@@ -123,7 +123,10 @@
 
 STATIC_URL = "static/"
 
-STATICFILES_DIRS = (BASE_DIR / "static",)
+STATICFILES_DIRS = [
+    BASE_DIR / "static",
+]
+# STATICFILES_DIRS = (BASE_DIR / "static",)
```
Y-Havryliv, line 129: why do you need it here?

templates/taxi/car_list.html
```diff
@@ -0,0 +1,23 @@
...
+       <tr>
+          <th>Id</th>
+          <th>Model</th>
+          <th>Manufacturer(name)</th>
+          <th>Manufacturer(country)</th>
+        </tr>
+      {% for car in car_list %}
+      <tr>
+        <td>{{ car.id }}</td>
+        <td><a href="{% url 'taxi:car-detail' pk=car.id %}">{{ car.model }}</a></td>
```
Y-Havryliv, line 16: In the car list set a link on the id field that leads to the car detail page.

Y-Havryliv (changes requested):

taxi_service/settings.py
alexzubkoff: This is the result of unsuccessful  experiments... 
Thanks!

Y-Havryliv (approved):
Well done!

PR #160, opened 2022-12-04
Develop

hai-le-di:
four
one
two
three

### DIFF
--- a/taxi/views.py
+++ b/taxi/views.py
@@ -1,4 +1,5 @@
 from django.shortcuts import render
+from django.views import generic
 
 from .models import Driver, Car, Manufacturer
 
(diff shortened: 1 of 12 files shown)


templates/includes/sidebar.html
```diff
@@ -1,6 +1,6 @@
 <ul class="sidebar-nav">
-  <li><a href="#">Home page</a></li>
-  <li><a href="#">Manufacturers</a></li>
-  <li><a href="#">Cars</a></li>
-  <li><a href="#">Drivers</a></li>
+  <li><a href="{% url "taxi:index" %}">Home page</a></li>
+  <li><a href="{% url "taxi:manufacturer-list" %}">Manufacturers</a></li>
+  <li><a href="{% url "taxi:car-list" %}">Cars</a></li>
+  <li><a href="{% url "taxi:driver-list" %}">Drivers</a></li>
```
OleksandrPozden, line 5: add a break line to all the HTML files in the end

OleksandrPozden (changes requested):

taxi/views.py
```diff
@@ -17,3 +18,36 @@ def index(request):
...
+class ManufacturerListView(generic.ListView):
+    model = Manufacturer
+    queryset = Manufacturer.objects.order_by("name")
+    template_name = "taxi/manufacturer_list.html"
+    paginate_by = 5
+
+
+class CarListView(generic.ListView):
+    model = Car
+    template_name = "taxi/car_list.html"
```
Serhii-Leonenko, line 32: Redundant field, you can remove it. In other similar cases too

taxi/views.py
```diff
@@ -17,3 +18,36 @@ def index(request):
...
+    paginate_by = 5
+    queryset = Car.objects.select_related("manufacturer")
+
+
+class CarDetailView(generic.DetailView):
+    model = Car
+    queryset = Car.objects.select_related("manufacturer").prefetch_related(
```
Serhii-Leonenko, line 39: You don't need it. Because there are no backward relations.

Serhii-Leonenko (changes requested):

taxi/views.py
Abnormaltype: what about it?)

Abnormaltype (changes requested):

PR #188, opened 2023-02-02
Solution Maksym Maksimov

MKeSiMu:


MKeSiMu:
**Manufacturers List**
image

**Cars List**
image

**Car Detail**
image

**Drivers List**
image

**Driver Detail**
image

taxi/views.py
```diff
@@ -17,3 +18,27 @@ def index(request):
     }
 
     return render(request, "taxi/index.html", context=context)
+
+
+class ManufacturerListView(generic.ListView):
```
katevasilenko, line 23: Please set context_object_name and template_name

taxi/views.py
```diff
@@ -17,3 +18,27 @@ def index(request):
...
+    model = Car
+
+
+class DriverListView(generic.ListView):
+    model = Driver
+    paginate_by = 5
+
+
+class DriverDetailView(generic.DetailView):
+    model = Driver
```
katevasilenko, line 44: Please set queryset with needed method in it

katevasilenko (changes requested):

taxi/views.py
MKeSiMu: Set context_object_name and template_name, but I don't really understand how in this case it would help. From video I understand, that context_object_name and template_name need set only in case when class base view split it in different way (for example "literaryformat_list", but you want it to be "literary_format_list"). In case with "manufacturer", class base view split it in correct way. Can you please explane to me, maybe I misunderstood some part. Thank you!

Grayder0152 (approved):

PR #189, opened 2023-02-02
Solution

Ihor-Ivashov:
manufact
Drivers
Driver
cars
car

### DIFF
--- a/taxi/views.py
+++ b/taxi/views.py
@@ -44,3 +44,4 @@ class DriverListView(generic.ListView):
 
 class DriverDetailView(generic.DetailView):
     model = Driver
+    queryset = Driver.objects.prefetch_related("cars__manufacturer")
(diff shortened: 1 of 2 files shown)


taxi/views.py
```diff
@@ -17,3 +18,29 @@ def index(request):
...
+class CarDetailView(generic.DetailView):
+    model = Car
+
+
+class DriverListView(generic.ListView):
+    model = Driver
+    paginate_by = 5
+
+
+class DriverDetailView(generic.DetailView):
```
ArsenPidhoretskyi, line 45: > In this view, you display information about cars of the driver. Optimize query: don't make a query for a manufacturer for each car (N+1 problem).

ArsenPidhoretskyi (changes requested):

Vasyl-Smutok (approved):
Good Job 👍

PR #214, opened 2023-03-13
Developer

manjustice:


AndereLion (approved):
OK!

taxi/urls.py
```diff
@@ -1,9 +1,34 @@
...
+        views.ManufacturerListView.as_view(),
+        name="manufacturer-list"
+    ),
+    path(
+        "cars/",
+        views.CarListView.as_view(),
+        name="car-list"
+    ),
+    path(
+        "cars/<int:pk>",
```
wQuelS, line 18: seems like you're missing `/` after `pk` in your `path`. Fix everywhere

wQuelS (changes requested):

taxi_service/urls.py
```diff
@@ -21,6 +21,6 @@
 
 urlpatterns = [
     path("admin/", admin.site.urls),
-    path("", include("taxi.urls", namespace="taxi")),
+    path("/", include("taxi.urls", namespace="taxi")),
```
Nattalli, line 24: but here it's unnecessary. Please, remove it 😄

PR #218, opened 2023-03-15
Solution

SafonovVladimir:


usrofgh (approved):

taxi/views.py
```diff
@@ -17,3 +18,30 @@ def index(request):
     }
 
     return render(request, "taxi/index.html", context=context)
+
+
+class ManufacturerListView(generic.ListView):
+    model = Manufacturer
+    template_name = "taxi/manufacturer_list.html"
+    context_object_name = "manufacturer_list"
+    queryset = Manufacturer.objects.all().order_by("name")
```
YuriiKindrat, line 27: no need to use all before order_by method

taxi/views.py
```diff
@@ -17,3 +18,30 @@ def index(request):
...
+    model = Car
+
+
+class DriverListView(generic.ListView):
+    model = Driver
+    paginate_by = 5
+
+
+class DriverDetailView(generic.DetailView):
+    model = Driver
```
YuriiKindrat, line 47: image

YuriiKindrat (changes requested):

taxi/views.py
SafonovVladimir: Queryset was added

taxi/views.py
SafonovVladimir: Fixed

PR #219, opened 2023-03-17
Solution

MaybeTI:
image
image
image
image
image
image

### DIFF
--- a/taxi/views.py
+++ b/taxi/views.py
@@ -1,4 +1,5 @@
 from django.shortcuts import render
+from django.views.generic import ListView, DetailView
 
 from .models import Driver, Car, Manufacturer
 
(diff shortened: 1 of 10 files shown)


yura-hudzovskyi (approved):

taxi/views.py
```diff
@@ -17,3 +18,34 @@ def index(request):
...
+    template_name = "taxi/manufacturer_list.html"
+    paginate_by = 5
+    ordering = "name"
+
+
+class CarListView(ListView):
+    model = Car
+    paginate_by = 5
+    queryset = Car.objects.all().select_related("manufacturer")
+    # template_name = "taxi/car_list.html"
```
YuriiKindrat, line 35: remove unnecessary comment here, also no need to use all with select related method

YuriiKindrat (changes requested):

taxi/views.py
```diff
@@ -32,20 +32,17 @@ class CarListView(ListView):
     model = Car
     paginate_by = 5
     queryset = Car.objects.all().select_related("manufacturer")
```
Nattalli, line 34: please, dont use `.all()` while using `.select_related()` or `.preferch_related()`. It's unnecessary and bad practice

taxi/views.py
```diff
@@ -32,20 +32,17 @@ class CarListView(ListView):
...
 class DriverListView(ListView):
     model = Driver
     paginate_by = 5
-    # template_name = "taxi/driver_list.html"
 
 
 class DriverDetailView(DetailView):
     model = Driver
-    # template_name = "taxi/driver_detail.html"
+    queryset = Driver.objects.all().prefetch_related("cars")
```
Nattalli, line 48: the same here

Nattalli (changes requested):

taxi/views.py
Nattalli: that is working for all additional calling. For example, you shouldn't use `.all()` if you use `filter()` and so on

PR #229, opened 2023-03-21
'Solution'

Morty67:


Morty67:
main page

Morty67:
list of manufacturers

Morty67:
list of cars

Morty67:
info about car

Morty67:
list of drivers

Morty67:
info about driver

taxi/views.py
```diff
@@ -17,3 +18,30 @@ def index(request):
...
+    model = Car
+
+
+class DriverListView(generic.ListView):
+    model = Driver
+    paginate_by = 5
+
+
+class DriverDetailView(generic.DetailView):
+    model = Driver
```
ZabFTFT, line 47: I think we have to fix N + 1 problem here due to readme

ZabFTFT (approved):
GJ!

taxi/views.py
```diff
@@ -17,3 +18,33 @@ def index(request):
...
 
     return render(request, "taxi/index.html", context=context)
+
+
+class ManufacturerListView(generic.ListView):
+    model = Manufacturer
+    template_name = "taxi/manufacturer_list.html"
+    context_object_name = "manufacturer_list"
+    paginate_by = 5
+    queryset = Manufacturer.objects.all().order_by("name")
```
wQuelS, line 28: no need to write `all()` here. As well as before `select_related` and `prefetch_related`. Fix everywhere

templates/taxi/index.html
wQuelS, line 1: Your screenshots are a bit confusing with contrast since only the "main" page is aligned in center:)
It is not the reason for reject, I guess, but maybe you will take this into consideration.

wQuelS (changes requested):

templates/taxi/index.html
Morty67: yes i know it was for educational purposes
next time without it ))

PR #230, opened 2023-03-21
solution

DavidGlusch:
home
manu
carlist
cardisplay
driverlist
driverdisplay

### DIFF
--- a/taxi/views.py
+++ b/taxi/views.py
@@ -30,7 +30,7 @@ class ManufacturerListView(generic.ListView):
 
 class CarListView(generic.ListView):
     model = Car
-    queryset = Car.objects.all().select_related("manufacturer")
+    queryset = Car.objects.select_related("manufacturer")
     paginate_by = 5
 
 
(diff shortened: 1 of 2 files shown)


SergiiMachulin (approved):
LGTM

taxi/views.py
```diff
@@ -17,3 +18,31 @@ def index(request):
...
+
+
+class DriverListView(generic.ListView):
+    model = Driver
+    paginate_by = 5
+
+
+class DriverDetailView(generic.DetailView):
+    model = Driver
+    queryset = Driver.objects.all().prefetch_related("cars")
```
YuriiKindrat, line 48: no need to use all before select/refetch related

templates/includes/pagination.html
```diff
@@ -0,0 +1,26 @@
...
+
+  {% if page_obj.has_previous %}
+    <li class="page-item">
+      <a href="?page={{ page_obj.previous_page_number }}" class="page-link">prev</a>
+    </li>
+
+  {% endif %}
+
+
+
```
YuriiKindrat, line 12: why do u need so many blank lines here?)

YuriiKindrat (changes requested):

PR #234, opened 2023-03-21
Solution

SergiiMachulin:
###Manufacturer List
manuf

###Car List
car_list

###Car detail
car_detail

###Driver List
driver_list

###Driver detail
driver_detail


amber-marichi (approved):
nicely done. but why dont you move all style from your html's to your custom static/.css file and then import it to base.html ?

SergiiMachulin:
I had  problems with .css file, could not even change background-color. That's why I decided do it through html

> nicely done. but why dont you move all style from your html's to your custom static/.css file and then import it to base.html ?

YuriiKindrat (approved):

PR #241, opened 2023-03-22
Solution

Litvinenko23:
Screenshot at Mar 22 15-24-23

Screenshot at Mar 22 15-26-03

### DIFF
--- a/taxi/views.py
+++ b/taxi/views.py
@@ -1,4 +1,6 @@
+from django.http import Http404
 from django.shortcuts import render
+from django.views import generic
 
 from .models import Driver, Car, Manufacturer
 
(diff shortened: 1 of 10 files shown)


templates/includes/pagination.html
```diff
@@ -0,0 +1,19 @@
...
+        <span class="page-link">{{ page_obj.number }} of {{ paginator.num_pages }}</span>
+      </li>
+
+      {% if page_obj.has_next %}
+        <li class="page-item">
+          <a href="?page={{ page_obj.next_page_number }}" class="page-link">next</a>
+        </li>
+      {% endif %}
+    </ul>
+  {% endif %}
```
ostapT, line 19: should be a blank line at the end of each html file

ostapT (changes requested):

taxi/views.py
```diff
@@ -17,3 +19,31 @@ def index(request):
     }
 
     return render(request, "taxi/index.html", context=context)
+
+
+class ManufacturerListView(generic.ListView):
+    model = Manufacturer
+    template_name = "taxi/manufacturer_list.html"
+    context_object_name = "manufacturer_list"
+    queryset = Manufacturer.objects.all().order_by("name")
```
YuriiKindrat, line 28: no need to use all with order_by method

YuriiKindrat (changes requested):

PR #242, opened 2023-03-22
 Solution

tytomyr:
image

image

image

image

image

### DIFF
--- a/taxi/views.py
+++ b/taxi/views.py
@@ -22,7 +22,7 @@ def index(request):
 
 class ManufacturerListView(generic.ListView):
     model = Manufacturer
-    queryset = Manufacturer.objects.all().order_by("name")
+    queryset = Manufacturer.objects.order_by("name")
     context_object_name = "manufacturer_list"
     paginate_by = 5
 
--- a/templates/taxi/car_detail.html
+++ b/templates/taxi/car_detail.html
@@ -1,6 +1,6 @@
 {% extends "base.html" %}
 {% block content %}
-  <h1>{{ car.model }}</h1>
+  <h1>{{ car.model }} ({{ car.manufacturer.name }}, {{ car.manufacturer.country }})</h1>
     <h3>
       Drivers:
       {% for driver in car.drivers.all %}


taxi/views.py
```diff
@@ -17,3 +18,34 @@ def index(request):
...
+    paginate_by = 5
+
+
+class CarDetailView(generic.DetailView):
+    model = Car
+
+
+class DriverListView(generic.ListView):
+    model = Driver
+    template_name = "taxi/driver_list.html"
```
ZhAlexR, line 43: "driver_list.html" is a default name for ListView with model name "Driver". It is not obligatory to rewrite template name here

taxi/views.py
```diff
@@ -17,3 +18,34 @@ def index(request):
...
+    paginate_by = 5
+
+
+class DriverDetailView(generic.DetailView):
+    model = Driver
+    template_name = "taxi/driver_detail.html"
```
ZhAlexR, line 49: The same here. "driver_detail.html" is a default name.

ZhAlexR (approved):

taxi/views.py
```diff
@@ -17,3 +18,34 @@ def index(request):
...
+    queryset = Driver.objects.prefetch_related("cars__manufacturer")
```
ZhAlexR, line 50: Check you field in `prefetch_releted("cars__manufacturer")` against to `prefetch_releted("cars")` in "Django Debug Tollbar", 
because  in my case 2 requests to database were done when `prefetch_releted("cars")`, and I have tested the variant with `prefetch_releted("cars__manufacturer")` - 3 requests were done.

taxi/urls.py
```diff
@@ -1,9 +1,29 @@
...
 
 urlpatterns = [
     path("", index, name="index"),
+    path("manufacturer/",
+         ManufacturerListView.as_view(),
+         name="manufacturer-list"),
+    path("cars/",
+         CarListView.as_view(),
+         name="car-list"),
+    path("car/<int:pk>/",
```
Y-Havryliv, line 18:
```suggestion
    path("cars/<int:pk>/",
```

taxi/urls.py
```diff
@@ -1,9 +1,29 @@
...
+         CarDetailView.as_view(),
+         name="car-detail"),
+    path("driver/",
```
Y-Havryliv, line 21:
```suggestion
    path("drivers/",
```

taxi/views.py
Y-Havryliv: agree

taxi/views.py
Y-Havryliv: I agree

taxi/views.py
Y-Havryliv: if you use car's manufacturer in driver detail template - then this is correct, if no - then change it

templates/taxi/car_list.html
```diff
@@ -0,0 +1,14 @@
+{% extends "base.html" %}
```
Y-Havryliv, line 1: Create templates for manufacturer list, car list, driver list. In these templates:
Create table with the information of each instance.

templates/taxi/driver_list.html
```diff
@@ -0,0 +1,21 @@
...
+{% block content %}
+  <h1>Drivers page</h1>
+
+  {% for driver in driver_list %}
+
+    <ul>
+
+      <li >
+        <a href={% url "taxi:driver-detail" pk=driver.id%}>
+          {{ driver.first_name }} {{ driver.last_name }}
```
Y-Havryliv, line 11: In the driver list set a link on the username field that leads to the driver detail page.

templates/taxi/car_detail.html
```diff
@@ -0,0 +1,13 @@
+{% extends "base.html" %}
+{% block content %}
+  <h1>{{ car.model }}</h1>
```
Y-Havryliv, line 3: Create car detail template:
place information about the manufacturer of the car (name, country);

Y-Havryliv (changes requested):
read README more thoroughly

taxi/views.py
```diff
@@ -17,3 +18,31 @@ def index(request):
     }
 
     return render(request, "taxi/index.html", context=context)
+
+
+class ManufacturerListView(generic.ListView):
+    model = Manufacturer
+    queryset = Manufacturer.objects.all().order_by("name")
```
YuriiKindrat, line 25: no need to use all before order by method

templates/taxi/car_detail.html
YuriiKindrat: u didn't fix that 
image

YuriiKindrat (approved):

PR #252, opened 2023-03-22
solution

Yulia-Mala:


Yulia-Mala:
Monosnap Taxi Service 2023-03-22 21-32-54
Monosnap Taxi Service 2023-03-22 21-32-43
Monosnap Taxi Service 2023-03-22 21-32-33
Monosnap Taxi Service 2023-03-22 21-32-19
Monosnap Taxi Service 2023-03-22 21-32-00

Danil228 (approved):

templates/taxi/car_list.html
```diff
@@ -0,0 +1,18 @@
...
+      <li>
+        <a href={% url "taxi:car-detail" pk=car.id %}>{{ car.id }}.</a>
+        {{ car.model }}
+      </li>
+    {% endfor %}
+    </ul>
+  {% else %}
+    <p>There are no cars..;(</p>
+  {% endif %}
+{% endblock %}
```
YuriiKindrat, line 18: no blank line at the end of file

YuriiKindrat (changes requested):

templates/taxi/car_list.html
Yulia-Mala: done, but could you pls tell me - why is it so important to add blank line at the end of file?)

wQuelS: Conventions, PEP8 if we are talking about `.py` files. But I guess your concern was more about `.html` and other files. Some reasons: VCS like Git track changes line by line. Git detects differences between lines. Its documentation says that if a file does not end with a newline character, adding a new line at the end will result in both the last line and the new line being marked as changed. Because of that some unnecessary noise in the diffs may appear, making reviewing process more difficult.
Another reason is more in a CS (Computer Science) scope. You can read on your own about POSIX compliance (Portable Operating System Interface).

wQuelS (approved):

PR #257, opened 2023-03-22
Solution

myroslavatrschk:
chrome_XQCKrZBD8g
chrome_ECU22quLZA
chrome_PQXtfrbnhA
chrome_HZCtaYQNYm
chrome_mXKdpUjVCY
chrome_sNoeUYUjef

### DIFF
--- a/taxi/urls.py
+++ b/taxi/urls.py
@@ -1,9 +1,29 @@
 from django.urls import path
 
-from .views import index
+from .views import (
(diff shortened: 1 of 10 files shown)


Y-Havryliv (changes requested):
Create templates for manufacturer list, car list, driver list. In these templates:
Create table with the information of each instance

PR #260, opened 2023-03-22
Solution

VladyslavCherkashyn:
image

image

image

image

image

### DIFF
--- a/taxi/views.py
+++ b/taxi/views.py
@@ -1,4 +1,5 @@
 from django.shortcuts import render
+from django.views import generic
 
 from .models import Driver, Car, Manufacturer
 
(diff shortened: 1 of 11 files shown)


taxi/views.py
```diff
@@ -17,3 +18,32 @@ def index(request):
...
+
+
+class DriverListView(generic.ListView):
+    model = Driver
+    paginate_by = 5
+
+
+class DriverDetailView(generic.DetailView):
+    model = Driver
+    queryset = Driver.objects.prefetch_related("cars__manufacturer")
```
YuriiKindrat, line 48: why do u need prefetch related to manufacturer if u are not using manufacturer info in ur driver detail view ?

PR #263, opened 2023-03-22
solution

AlessaQ:


taxi/views.py
```diff
@@ -17,3 +18,31 @@ def index(request):
     }
 
     return render(request, "taxi/index.html", context=context)
+
+
+class ManufacturerListView(generic.ListView):
+    model = Manufacturer
+    context_object_name = "manufacturer_list"
+    queryset = Manufacturer.objects.all().order_by("name")
```
YuriiKindrat, line 26: no need to use all method here

taxi/views.py
```diff
@@ -17,3 +18,31 @@ def index(request):
...
+    template_name = "taxi/manufacturer_list.html"
+    paginate_by = 5
+
+
+class CarListView(generic.ListView):
+    model = Car
+    queryset = Car.objects.all().select_related("manufacturer")
```
YuriiKindrat, line 33: same

YuriiKindrat (changes requested):

taxi/views.py
AlessaQ: Fixed

templates/taxi/driver_detail.html
```diff
@@ -0,0 +1,21 @@
...
+  <h1>Call sign: {{ driver.username }}</h1>
+  <hr>
+  <p><strong>License number:</strong> {{ driver.license_number }}</p>
+  <p><strong>Name: </strong>{{ driver.first_name }} {{ driver.last_name }}</p>
+
+  <div class="ml-3">
+  <h3>Driver's cars:</h3>
+  {% for car in driver.cars.all %}
+    <ul>
+      <a href="{% url 'taxi:car-detail' pk=car.id %}">{{ car.model }}</a>
```
wQuelS, line 13: I guess that not only model, but also the `manufacturer` info should be shown here

wQuelS (changes requested):

templates/taxi/driver_detail.html
AlessaQ: image

PR #264, opened 2023-03-23
Solution

Zenionoi:


templates/includes/pagination.html
```diff
@@ -0,0 +1,20 @@
...
+    </li>
+
+    {% if page_obj.has_next %}
+      <li class="page-item">
+        <a href="?page= {{ page_obj.next_page_number }}" class="page-link">next</a>
+      </li>
+
+    {% endif %}
+  </ul>
+{% endif %}
```
YuriiKindrat, line 20: no blank line at the end of file

taxi/views.py
```diff
@@ -17,3 +18,31 @@ def index(request):
...
+
+
+class DriverListView(generic.ListView):
+    model = Driver
+    paginate_by = 5
+
+
+class DriverDetailView(generic.DetailView):
+    model = Driver
+    queryset = Driver.objects.prefetch_related("cars__manufacturer")
```
YuriiKindrat, line 48: do u use manufacturer info in ur driver detail template anywhere?

YuriiKindrat (changes requested):

taxi/views.py
Zenionoi: No, I thought about the possibility of displaying information about car's manufacturer
Maybe `Model: model (manufacturer)` or something similar, but many of `car.model` already have mention of manufacturer

taxi/views.py
```diff
@@ -17,3 +18,30 @@ def index(request):
...
+    model = Car
+
+
+class DriverListView(generic.ListView):
+    model = Driver
+    paginate_by = 5
+
+
+class DriverDetailView(generic.DetailView):
+    model = Driver
```
YuriiKindrat, line 47: image
u are still using car info in ur template, so fix n+1 problem

YuriiKindrat (changes requested):

PR #316, opened 2023-06-05
solution

Bloodviel:

### DIFF
--- a/taxi/urls.py
+++ b/taxi/urls.py
@@ -1,9 +1,41 @@
 from django.urls import path
 
-from .views import index
+from .views import (
+    index,
+    CarListView,
+    CarDetailView,
+    DriverListView,
+    DriverDetailView,
+    ManufacturerListView
+)
 
 urlpatterns = [
     path("", index, name="index"),
+    path(
+        "Manufacturer-list/",
+        ManufacturerListView.as_view(),
+        name="manufacturer-list"
+    ),
+    path(
+        "Car-list/",
+        CarListView.as_view(),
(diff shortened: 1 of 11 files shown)


taxi/urls.py
```diff
@@ -1,9 +1,41 @@
...
+        ManufacturerListView.as_view(),
+        name="manufacturer-list"
+    ),
+    path(
+        "Car-list/",
+        CarListView.as_view(),
+        name="car-list"
+    ),
+    path(
+        "Car-detail/<int:pk>",
```
vasylhnatiuk, line 25:
```suggestion
        "car-detail/<int:pk>/",
```

taxi/urls.py
```diff
@@ -1,9 +1,41 @@
...
+    CarDetailView,
+    DriverListView,
+    DriverDetailView,
+    ManufacturerListView
+)
 
 urlpatterns = [
     path("", index, name="index"),
+    path(
+        "Manufacturer-list/",
```
vasylhnatiuk, line 15:
```suggestion
        "manufacturers/",
```

vasylhnatiuk: Use lower case letters and without a 'list', but the word in plural. Here and all of the code

taxi/urls.py
vasylhnatiuk: add / in the end here and all of the code

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):
Good job

PR #337, opened 2023-06-26
solution

OksanaAdamchuk:

### DIFF
--- a/taxi/urls.py
+++ b/taxi/urls.py
@@ -1,9 +1,23 @@
 from django.urls import path
 
-from .views import index
+from .views import CarDetailView, CarListView, DriverDetailView, \
+    DriverListView, ManufacturerListView, index
 
 urlpatterns = [
     path("", index, name="index"),
+    path(
+        "manufacturers/",
+        ManufacturerListView.as_view(),
+        name="manufacturer-list"
+    ),
+    path("cars/", CarListView.as_view(), name="car-list"),
+    path("cars/<int:pk>/", CarDetailView.as_view(), name="car-detail"),
+    path("drivers/", DriverListView.as_view(), name="driver-list"),
+    path(
+        "drivers/<int:pk>/",
+        DriverDetailView.as_view(),
+        name="driver-detail"
+    ),
 ]
 
 app_name = "taxi"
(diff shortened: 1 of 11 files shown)


taxi/views.py
```diff
@@ -17,3 +18,34 @@ def index(request):
...
+    model = Car
+    context_object_name = "car_list"
+    paginate_by = 5
+    template_name = "taxi/car_list.html"
+    queryset = Car.objects.all().select_related("manufacturer")
+    queryset = Car.objects.all().prefetch_related("drivers")
+
+
+class CarDetailView(DetailView):
+    model = Car
```
ArsenPidhoretskyi, line 41: add select related and prefetch related

taxi/views.py
```diff
@@ -17,3 +18,34 @@ def index(request):
...
+    queryset = Manufacturer.objects.all().order_by("name")
+
+
+class CarListView(ListView):
+    model = Car
+    context_object_name = "car_list"
+    paginate_by = 5
+    template_name = "taxi/car_list.html"
+    queryset = Car.objects.all().select_related("manufacturer")
+    queryset = Car.objects.all().prefetch_related("drivers")
```
ArsenPidhoretskyi, line 37: As a result you will have only last queryset. Do:
```suggestion
    queryset = Car.objects.all().select_related("manufacturer").prefetch_related("drivers")
```

ArsenPidhoretskyi (changes requested):

taxi/urls.py
```diff
@@ -1,9 +1,23 @@
 from django.urls import path
 
-from .views import index
+from .views import CarDetailView, CarListView, DriverDetailView, \
+    DriverListView, ManufacturerListView, index
```
Grayder0152, line 5: use `()` instead of `\`

taxi/views.py
```diff
@@ -17,3 +18,36 @@ def index(request):
...
+    queryset = Manufacturer.objects.all().order_by("name")
+
+
+class CarListView(ListView):
+    model = Car
+    context_object_name = "car_list"
+    paginate_by = 5
+    template_name = "taxi/car_list.html"
+    queryset = Car.objects.all().select_related("manufacturer"). \
+        prefetch_related("drivers")
```
Grayder0152, line 37: the same as above

Grayder0152 (changes requested):

PR #348, opened 2023-07-06
Solution

gleblitvinenko:
image
image
image
image
image
image

### DIFF
--- a/taxi/urls.py
+++ b/taxi/urls.py
@@ -21,7 +21,7 @@ urlpatterns = [
         name="car-list",
     ),
     path(
-        "cars/<int:pk>",
+        "cars/<int:pk>/",
         CarDetailView.as_view(),
         name="car-detail",
     ),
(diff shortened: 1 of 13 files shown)


taxi/urls.py
```diff
@@ -1,9 +1,40 @@
...
+        "manufacturers/", ManufacturerListView.as_view(),
+        name="manufacturer-list"
+    ),
+    path(
+        "cars/",
+        CarListView.as_view(),
+        name="car-list",
+    ),
+    path(
+        "cars/<int:pk>",
```
ArsenPidhoretskyi, line 24: Path must end at /

taxi/views.py
```diff
@@ -17,3 +18,31 @@ def index(request):
...
+
+
+class CarListView(generic.ListView):
+    model = Car
+    paginate_by = 5
+    queryset = Car.objects.select_related("manufacturer")
+
+
+class CarDetailView(generic.DetailView):
+    model = Car
```
ArsenPidhoretskyi, line 38: Add select_related and prefetch_related to optimize query. Because inside html you use drivers and manufacturer

ArsenPidhoretskyi (changes requested):

vsmutok (approved):
Great 👍

PR #352, opened 2023-07-12
solution

oleksiikolii:


taxi/urls.py
```diff
@@ -1,9 +1,35 @@
...
+    path(
+        "drivers/",
+        DriverListView.as_view(),
+        name="driver-list",
+    ),
+    path(
+        "drivers/<int:pk>",
+        DriverDetailView.as_view(),
+        name="driver-detail"
+    )
```
YuriiKindrat, line 32: add `/` at the end of all urls )

taxi/views.py
```diff
@@ -17,3 +18,30 @@ def index(request):
...
+    model = Manufacturer
+    context_object_name = "manufacturer_list"
+    queryset = Manufacturer.objects.all().order_by("name")
+    template_name = "taxi/manufacturer_list.html"
+    paginate_by = 5
+
+
+class CarListView(generic.ListView):
+    model = Car
+    queryset = Car.objects.all().select_related("manufacturer")
```
YuriiKindrat, line 33: why do u need all method here ?

YuriiKindrat (changes requested):

taxi/views.py
```diff
@@ -17,3 +18,30 @@ def index(request):
     }
 
     return render(request, "taxi/index.html", context=context)
+
+
+class ManufacturerListView(generic.ListView):
+    model = Manufacturer
+    context_object_name = "manufacturer_list"
+    queryset = Manufacturer.objects.all().order_by("name")
```
YuriiKindrat, line 26: same is here, will this work without all method ?

taxi/views.py
```diff
@@ -17,3 +18,30 @@ def index(request):
...
+class CarDetailView(generic.DetailView):
+    model = Car
+
+
+class DriverListView(generic.ListView):
+    model = Driver
+    paginate_by = 5
+
+
+class DriverDetailView(generic.DetailView):
```
YuriiKindrat, line 46: what about N+1 problem here?

YuriiKindrat (approved):

PR #353, opened 2023-07-12
Solution

diana-shyrokikh:
Manufacturer List:
image

Car List:
image

Car Detail:
image

Driver List:
image

Driver Detail:
image

### DIFF
--- a/taxi/views.py
+++ b/taxi/views.py
@@ -1,4 +1,5 @@
 from django.shortcuts import render
+from django.views import generic
 
 from .models import Driver, Car, Manufacturer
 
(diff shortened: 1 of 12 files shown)


taxi/views.py
```diff
@@ -17,3 +18,32 @@ def index(request):
...
+
+
+class DriverListView(generic.ListView):
+    model = Driver
+    paginate_by = 5
+
+
+class DriverDetailView(generic.DetailView):
+    model = Driver
+    queryset = Driver.objects.all().prefetch_related("cars")
```
YuriiKindrat, line 48: why do u need all method here ?

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #359, opened 2023-07-13
Solution

Kirontiko:
Home page

image

Manufacturers

image

Cars

image

Car detail info example
image

Drivers

image

Driver detail info example

image

### DIFF
--- a/taxi/urls.py
+++ b/taxi/urls.py
@@ -16,13 +16,13 @@ urlpatterns = [
     path("cars/",
          CarListView.as_view(),
          name="car-list"),
-    path("cars/<int:pk>",
+    path("cars/<int:pk>/",
          CarDetailView.as_view(),
          name="car-detail"),
     path("drivers/",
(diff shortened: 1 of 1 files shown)


taxi/urls.py
```diff
@@ -1,9 +1,30 @@
...
+    path("cars/",
+         CarListView.as_view(),
+         name="car-list"),
+    path("cars/<int:pk>",
+         CarDetailView.as_view(),
+         name="car-detail"),
+    path("drivers/",
+         DriverListView.as_view(),
+         name="driver-list"),
+    path("drivers/<int:pk>",
```
YuriiKindrat, line 25: link should end with `/`

YuriiKindrat (changes requested):

taxi/urls.py
Kirontiko: oh, will fix it, sorry(

PR #363, opened 2023-07-13
Develop

IhorVoskoboinikov:
All cars menu
All drivers menu
All manufacturers menu
Detail info about car
Detail info about driver
Home page taxi service

### DIFF
--- a/taxi/views.py
+++ b/taxi/views.py
@@ -1,4 +1,5 @@
 from django.shortcuts import render
+from django.views import generic
 
 from .models import Driver, Car, Manufacturer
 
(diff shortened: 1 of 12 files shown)


taxi/views.py
```diff
@@ -17,3 +18,29 @@ def index(request):
     }
 
     return render(request, "taxi/index.html", context=context)
+
+
+class ManufacturerListView(generic.ListView):
+    model = Manufacturer
+    queryset = Manufacturer.objects.order_by("name")
+    paginate_by = 5
```
YuriiKindrat, line 26: it seems to me, that according to the task u should add 2 more attrs to this view )

templates/taxi/car_detail.html
```diff
@@ -0,0 +1,18 @@
...
+  <lu>
+    <li>{{ car.manufacturer.name }}</li>
+    <li>{{ car.manufacturer.country }}</li>
+  </lu>
+  <h4>Information about all drivers:</h4>
+  {% for drivers in  car.drivers.all %}
+    <lu>
+      <li>{{ drivers.first_name }}</li>
+    </lu>
+  {% endfor %}
```
YuriiKindrat, line 17: what if car won't have related drivers ? add message for such case and for similar ones

YuriiKindrat (changes requested):

PR #364, opened 2023-07-13
My solution

kovaliskoveronika:
car_detail_page
carlist_page
driver_detail_page
drivers_list_page
home_page
manf_page

### DIFF
--- a/taxi/views.py
+++ b/taxi/views.py
@@ -1,6 +1,7 @@
 from django.shortcuts import render
 
 from .models import Driver, Car, Manufacturer
+from django.views import generic
 
 
 def index(request):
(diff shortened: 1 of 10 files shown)


taxi/views.py
```diff
@@ -17,3 +18,30 @@ def index(request):
...
+    context_object_name = "manufacturer_list"
+    queryset = Manufacturer.objects.order_by("name")
+    template_name = "taxi/manufacturer_list.html"
+    paginate_by = 5
+
+
+class CarListView(generic.ListView):
+    model = Car
+    paginate_by = 5
+    queryset = Car.objects.all().select_related("manufacturer")
```
LiudmylaKulinchenko, line 34: No need to use `all` here

templates/taxi/driver_detail.html
```diff
@@ -0,0 +1,15 @@
+{% extends "base.html" %}
+
+{% block content %}
+  <p class="font-weight-bold">{{ driver.first_name }} {{ driver.last_name }}</p>
+  <ul>
+    {% for car in driver.cars.all %}
```
LiudmylaKulinchenko, line 6: You're getting car from driver, change your view to resolve n+1 problem

LiudmylaKulinchenko (changes requested):

taxi/views.py
```diff
@@ -17,3 +18,32 @@ def index(request):
...
+
+
+class CarDetailView(generic.DetailView):
+    model = Car
+
+
+class DriverListView(generic.ListView):
+    model = Driver
+    paginate_by = 5
+    queryset = Driver.objects.prefetch_related("cars")
```
YuriiKindrat, line 44: why do u need to use prefetch related here ?

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #365, opened 2023-07-13
Develop ot

OlhaTryhub:
image

image

image

image

image

image


static/css/styles.css
```diff
@@ -1,3 +1,31 @@
...
+table {
+    max-width: 700px;
+    min-width: 350px;
+    box-shadow: 10px 10px lightblue;
+    background-color: aliceblue;
+}
+
+.sidebar-nav {
+    background-color: aliceblue
+}
```
YuriiKindrat, line 31: no blank line at the end of file(

templates/taxi/car_detail.html
```diff
@@ -0,0 +1,36 @@
...
+          {% for driver in car.drivers.all %}
+            <tr>
+              <td>
+                <a href="{% url 'taxi:driver-detail' pk=driver.id%}">
+                  {{ driver.last_name }} {{ driver.first_name }}
+                </a>
+              </td>
+              <td>{{ driver.license_number }}</td>
+            </tr>
+          {% endfor %}
```
YuriiKindrat, line 31: what if there won't be drivers ? Add message for this case, the same is for similar cases

YuriiKindrat (changes requested):

PR #367, opened 2023-07-13
solution

mariiahorbova:
image
image
image
image
image
image


Oleg-s-s (approved):
If you use env its better if you create some .example file with fields and example filling.
f.e. settings.ini.example or .env.example and push it with pr. I see you set a default value for consts that use value from env but if there will be a lot of such consts it will be difficult to find every single one for creating personal .env .ini

PR #376, opened 2023-07-17
Solution

Po-SvinSkipodla:
image
image
image
image
image
image

### DIFF
--- a/taxi/views.py
+++ b/taxi/views.py
@@ -1,3 +1,4 @@
+from django.views import generic
 from django.shortcuts import render
 
 from .models import Driver, Car, Manufacturer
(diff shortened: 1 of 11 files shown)


taxi/views.py
```diff
@@ -17,3 +18,31 @@ def index(request):
...
+
+
+class DriverListView(generic.ListView):
+    model = Driver
+    paginate_by = 5
+
+
+class DriverDetailView(generic.DetailView):
+    model = Driver
+    queryset = Driver.objects.all().prefetch_related("cars")
```
YuriiKindrat, line 48: why do u need "all()" methods here?

templates/taxi/car_detail.html
```diff
@@ -0,0 +1,14 @@
...
+  <h3>Manufacturer: {{ car.manufacturer.name }}</h3>
+  <h3>Country: {{ car.manufacturer.country }}</h3>
+  <h3>Drivers: </h3>
+  <ul class="list">
+    {% for driver in car.drivers.all %}
+      <li>
+        {{ driver.username }}
+      </li>
+    {%  endfor %}
+  </ul>
```
YuriiKindrat, line 13: what if there won't be related drivers ? it seems to me it is better to add message for such cases
fix this here and in similar cases

YuriiKindrat (changes requested):

templates/taxi/car_detail.html
```diff
@@ -5,10 +5,12 @@ <h3>Manufacturer: {{ car.manufacturer.name }}</h3>
...
-        {{ driver.username }}
-      </li>
-    {%  endfor %}
+    {% if car.drivers.all %}
+      {% for driver in car.drivers.all %}
+        <li>
+          {{ driver.username }}
+        </li>
+      {%  endfor %}
+    {% endif %}
```
YuriiKindrat, line 14: u can add some text for such case 
f.e. "Car has no owners" or smth like that

YuriiKindrat (changes requested):

PR #391, opened 2023-08-15
py-taxi-service-class-based-views

muriakula:
driver 4
driver list
car 1
car list
manufacturer list
home page


YukihiroSM (approved):
Looks good to me!

PR #409, opened 2023-09-18
Develop

eduardhabryd:
Home page:

image

Manufacturers:

image

Cars:

image

Drivers:

image

Single Car:

image

Single Driver:

image


OlegatorLE (approved):
Good Job!

taxi/views.py
```diff
@@ -17,3 +18,37 @@ def index(request):
...
+class DriverListView(ListView):
+    model = Driver
+    paginate_by = 5
+    context_object_name = "driver_list"
+    template_name = "taxi/driver_list.html"
+
+
+class DriverDetailView(DetailView):
+    model = Driver
+    queryset = Driver.objects.prefetch_related("cars", "cars__manufacturer")
```
fsocie7y: i think cars_manufacturer is redundant

fsocie7y (changes requested):

taxi/views.py
```diff
@@ -17,3 +18,37 @@ def index(request):
     }
 
     return render(request, "taxi/index.html", context=context)
+
+
+class ManufacturerListView(ListView):
+    model = Manufacturer
+    context_object_name = "manufacturer_list"
```
fsocie7y: u have redundant attributes in views
read task description

taxi/views.py
```diff
@@ -17,3 +18,37 @@ def index(request):
...
+
+class CarListView(ListView):
+    model = Car
+    context_object_name = "car_list"
+    queryset = Car.objects.order_by("model").select_related("manufacturer")
+    template_name = "taxi/car_list.html"
+    paginate_by = 5
+
+
+class CarDetailView(DetailView):
```
fsocie7y, line 35: u need optimize ur query `couse want get access to driver in ur template

templates/taxi/car_list.html
```diff
@@ -0,0 +1,15 @@
+{% extends "base.html" %}
```
fsocie7y, line 1: also not everywhere in the templates  use get_absolut_url which  implemented in the models (it was possible in the list templates)

fsocie7y (changes requested):

taxi/views.py
```diff
@@ -17,3 +18,30 @@ def index(request):
...
+
+class CarListView(ListView):
+    model = Car
+    queryset = Car.objects.order_by("model").select_related("manufacturer")
+    paginate_by = 5
+
+
+class CarDetailView(DetailView):
+    model = Car
+    queryset = Car.objects.prefetch_related("drivers")
```
fsocie7y, line 37: u also need to opimize query for manufacturer, `couse u use it in template

fsocie7y (changes requested):

taxi/views.py
eduardhabryd: > u also need to opimize query for manufacturer, `couse u use it in template

Every car has only one manufacturer, so there won't be N + 1 problem and, in my opinion, there's no need to optimize this query. If i'm wrong, please let me know.

fsocie7y (approved):

PR #413, opened 2023-09-20
Solution

nyanplague:
Screenshot 2023-09-20 at 14 43 33
Screenshot 2023-09-20 at 14 44 22
Screenshot 2023-09-20 at 14 44 45
Screenshot 2023-09-20 at 14 44 33
Screenshot 2023-09-20 at 14 44 38


taxi/views.py
```diff
@@ -17,3 +18,31 @@ def index(request):
     }
 
     return render(request, "taxi/index.html", context=context)
+
+
+class ManufacturerListView(generic.ListView):
+    model = Manufacturer
+    context_object_name = "manufacturer_list"
```
lgSilay, line 25: I'm not sure about is it nessesary to redefine "context_object_name" as "manufacturer_list", because "context_object_name" created by default as <"class__name"_list> (the same with "template_name"). How I`m understand, we should redefine  "context_object_name" if we have additional underscore("_") in our templates (for example: literary_format_list).

On the other hand, we have condition in README.md: "Set the appropriate context variable name, that you will use in the template."
It was not so clear what does is mean? Should we create another, not default name or what?:( 

After all, in other classes you did not change "context_object_name" and "template_name" and everything works perfectly.

lgSilay (approved):
Looks very nice!)

taxi/views.py
nyanplague: @lgSilay I wrote it just because of condition in Readme, I think it's great question for QnA :)

Serhii-Leonenko (changes requested):

taxi/views.py
```diff
@@ -17,3 +18,29 @@ def index(request):
...
+
+
+class CarListView(generic.ListView):
+    model = Car
+    queryset = Car.objects.select_related("manufacturer")
+    paginate_by = 5
+
+
+class CarDetailView(generic.DetailView):
+    model = Car
```
Serhii-Leonenko, line 35: Optimize query according to your template

taxi/views.py
```diff
@@ -17,3 +18,29 @@ def index(request):
...
+
+class ManufacturerListView(generic.ListView):
+    model = Manufacturer
+    queryset = Manufacturer.objects.order_by("name")
+    paginate_by = 5
+
+
+class CarListView(generic.ListView):
+    model = Car
+    queryset = Car.objects.select_related("manufacturer")
```
Serhii-Leonenko: Do you access manufacturer in your template?

templates/taxi/driver_detail.html
```diff
@@ -0,0 +1,13 @@
+{% extends 'base.html' %}
+{% block content %}
+  <h1>Driver {{ driver.first_name }}, {{ driver.last_name }}</h1>
+  <p>{{ car.manufacturer.name }}</p>
```
Serhii-Leonenko: how do you access the car before iteration through cars?)

Serhii-Leonenko (changes requested):

Serhii-Leonenko (approved):

PR #429, opened 2023-09-22
generics views eaten

maxkatkalov:
I deployed my solution [here](https://test5.techone.pp.ua/) because screenshots, not so comfortable to watch.


taxi/views.py
```diff
@@ -17,3 +19,56 @@ def index(request):
     }
 
     return render(request, "taxi/index.html", context=context)
+
+
+class ManufacturerListView(ListView):
+    model = Manufacturer
+    queryset = Manufacturer.objects.all().order_by("name")
```
katryana, line 26: It could be simpler, like `Manufacturer.objects.order_by("name")`

katryana (approved):
Looks cool!

Kroll410 (approved):

PR #470, opened 2023-11-14
Solution

prochigor:
image
image
image
image
image
image

### DIFF
--- a/taxi/views.py
+++ b/taxi/views.py
@@ -18,8 +18,7 @@ def index(request):
 
 class ManufacturerListView(generic.ListView):
     model = Manufacturer
-    queryset = Manufacturer.objects.all()
-    ordering = ["name"]
+    queryset = Manufacturer.objects.order_by("name")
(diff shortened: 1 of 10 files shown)


vshvanska (approved):

taxi/views.py
```diff
@@ -13,3 +14,30 @@ def index(request):
     }
 
     return render(request, "taxi/index.html", context=context)
+
+
+class ManufacturerListView(generic.ListView):
+    model = Manufacturer
+    queryset = Manufacturer.objects.all()
+    ordering = ["name"]
```
LiudmylaKulinchenko: You can also use `queryset = Manufacturer.objects.order_by("name")`

LiudmylaKulinchenko (approved):
Well done ☀

taxi/views.py
prochigor: I know but wasn't sure how is better)

PR #492, opened 2023-11-16
Solution

skoval7:
django3 1
django3 2

### DIFF
--- a/taxi/views.py
+++ b/taxi/views.py
@@ -1,5 +1,5 @@
 from django.shortcuts import render
-
+from django.views.generic import ListView, DetailView
 from taxi.models import Driver, Car, Manufacturer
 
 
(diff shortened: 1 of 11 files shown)


taxi/views.py
```diff
@@ -13,3 +13,34 @@ def index(request):
     }
 
     return render(request, "taxi/index.html", context=context)
+
+
+class ManufacturerListView(ListView):
+    model = Manufacturer
+    context_object_name = "manufacturer_list"
```
LiudmylaKulinchenko: it is same by default, you can remove this line. Same everywhere

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Looks great! 🎈

PR #493, opened 2023-11-16
Implement py-taxi-service-class-based-views

SevKrok:
image
image
image
image
image
image
image

### DIFF
--- a/taxi/views.py
+++ b/taxi/views.py
@@ -18,7 +18,7 @@ def index(request):
 
 class ManufacturerListView(generic.ListView):
     model = Manufacturer
-    queryset = Manufacturer.objects.all().order_by("name")
+    queryset = Manufacturer.objects.order_by("name")
     paginate_by = 5
 
 
--- a/taxi/views.py
+++ b/taxi/views.py
@@ -1,4 +1,5 @@
 from django.shortcuts import render
+from django.views import generic
 
 from taxi.models import Driver, Car, Manufacturer
 
@@ -13,3 +14,29 @@ def index(request):
     }
 
     return render(request, "taxi/index.html", context=context)
+
+
+class ManufacturerListView(generic.ListView):
+    model = Manufacturer
+    queryset = Manufacturer.objects.all().order_by("name")
+    paginate_by = 5
+
+
+class CarListView(generic.ListView):
+    model = Car
+    queryset = Car.objects.select_related("manufacturer")
+    paginate_by = 5
+
+
+class CarDetailView(generic.DetailView):
+    model = Car
+
+
+class DriverListView(generic.ListView):
+    model = Driver
+    paginate_by = 5
+
+
+class DriverDetailView(generic.DetailView):
+    model = Driver
+    queryset = Driver.objects.prefetch_related("cars", "cars__manufacturer")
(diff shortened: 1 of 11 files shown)


kostomeister (approved):
good jobs

taxi/views.py
```diff
@@ -13,3 +14,29 @@ def index(request):
     }
 
     return render(request, "taxi/index.html", context=context)
+
+
+class ManufacturerListView(generic.ListView):
+    model = Manufacturer
+    queryset = Manufacturer.objects.all().order_by("name")
```
LiudmylaKulinchenko: this code will work the same without `.all()`

LiudmylaKulinchenko (changes requested):

taxi/views.py
```diff
@@ -13,3 +14,29 @@ def index(request):
...
+
+class ManufacturerListView(generic.ListView):
+    model = Manufacturer
+    queryset = Manufacturer.objects.order_by("name")
+    paginate_by = 5
+
+
+class CarListView(generic.ListView):
+    model = Car
+    queryset = Car.objects.select_related("manufacturer")
```
fsocie7y: in ur template u use Drivers but here queryset do not optimized
same in other cases too

taxi/views.py
```diff
@@ -13,3 +14,29 @@ def index(request):
...
+    paginate_by = 5
+
+
+class CarDetailView(generic.DetailView):
```
fsocie7y, line 33: here u also use Drivers in template

taxi/views.py
```diff
@@ -13,3 +14,29 @@ def index(request):
...
+    model = Car
+
+
+class DriverListView(generic.ListView):
+    model = Driver
+    paginate_by = 5
+
```
fsocie7y, line 44: and here u use Cars in template, but again do not optimize query

fsocie7y (changes requested):

taxi/views.py
```diff
@@ -13,3 +14,31 @@ def index(request):
...
+class ManufacturerListView(generic.ListView):
+    model = Manufacturer
+    queryset = Manufacturer.objects.prefetch_related("cars").order_by("name")
+    paginate_by = 5
+
+
+class CarListView(generic.ListView):
+    model = Car
+    queryset = Car.objects.prefetch_related("manufacturer", "drivers")
```
wQuelS: Now, your approach would result in separate queries for both `manufacturer` and `drivers`. While it would still reduce the overall number of queries compared to not using any optimization, it is less efficient than using select_related for the ForeignKey relationship because it would not take advantage of the possible SQL join for the manufacturer field, leading to an extra query that could have been avoided.

Consider using `select` for manufacturer and `prefetch` for drivers by chaining those methods. Also, you may play a bit and compare number of queries

taxi/views.py
```diff
@@ -13,3 +14,31 @@ def index(request):
...
+    paginate_by = 5
+
+
+class CarDetailView(generic.DetailView):
+    model = Car
+    queryset = Car.objects.prefetch_related("manufacturer", "drivers")
```
wQuelS: same

wQuelS (changes requested):
GJ overall, but left 2 comments so you will be able to dig deeper into query optimization:)

wQuelS (approved):
Well done!

PR #498, opened 2023-11-20
sloutione

oleksandr-pupchenko:


taxi/views.py
```diff
@@ -1,15 +1,41 @@
...
+class DriverListView(generic.ListView):
+    paginate_by = 5
+    model = Driver
+
+
+class CarDetailView(generic.DetailView):
+    model = Car
+
+
+class DriverDetailView(generic.DetailView):
```
Oxbay, line 39: u need to optimize query (prefetch_related)

Oxbay (changes requested):
u need to optimize query (prefetch_related)

LiudmylaKulinchenko (approved):
Good job 🌼

PR #503, opened 2023-11-26
Solution

Sparix:
Manufacturer-list.html
image

cars-list.html
image

drivers-list.html
image

car-detail.html
image

driver-detail.html
image


ArsenPidhoretskyi (approved):
Well done!

PR #547, opened 2024-01-20
Solution

Lebwa1337:
image
image
image
image
image
image


taxi/urls.py
```diff
@@ -1,9 +1,25 @@
...
     path("", index, name="index"),
+    path(
+        "manufacturers/",
+        ManufacturerListView.as_view(),
+        name="manufacturer-list"
+    ),
+    path("cars/", CarListView.as_view(), name="car-list"),
+    path("cars/<int:pk>", CarDetailView.as_view(), name="car-detail"),
+    path("drivers/", DriverListView.as_view(), name="driver-list"),
+    path("drivers/<int:pk>", DriverDetailView.as_view(), name="driver-detail"),
```
pavlejviki: all path should end with a /

pavlejviki (changes requested):

taxi/urls.py
```diff
@@ -1,9 +1,29 @@
 from django.urls import path
 
-from .views import index
+from .views import (
```
viktoria-rybenchuk, line 3: use absolute path instead of relative

viktoria-rybenchuk (approved):
Good job! 
Read recommendation

PR #558, opened 2024-02-05
Solution

nicksetrakov:
Manufacturer_list
Car_list
Car_detail
Driver_list
Driver_detail

### DIFF
--- a/taxi/urls.py
+++ b/taxi/urls.py
@@ -1,7 +1,12 @@
 from django.urls import path
 
-from .views import index, ManufacturerListView, CarListView,\
-    CarDetailView, DriverListView, DriverDetailView
+from .views import (
+    index,
+    ManufacturerListView,
+    CarListView,
+    CarDetailView, DriverListView,
+    DriverDetailView
+)
 
 urlpatterns = [
     path("", index, name="index"),
--- a/taxi/urls.py
+++ b/taxi/urls.py
@@ -1,9 +1,23 @@
 from django.urls import path
 
-from .views import index
+from .views import index, ManufacturerListView, CarListView,\
+    CarDetailView, DriverListView, DriverDetailView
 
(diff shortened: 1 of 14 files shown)


taxi/urls.py
```diff
@@ -1,9 +1,23 @@
 from django.urls import path
 
-from .views import index
+from .views import index, ManufacturerListView, CarListView,\
```
fsocie7y: Do not use line breakers, better use ( ) instead. This will be bestpractice ^_^

taxi/views.py
```diff
@@ -13,3 +15,30 @@ def index(request):
...
+
+class ManufacturerListView(generic.ListView):
+    model = Manufacturer
+    queryset = Manufacturer.objects.order_by("name")
+    paginate_by = 5
+
+
+class CarListView(generic.ListView):
+    model = Car
+    queryset = Car.objects.select_related("manufacturer").order_by("id")
```
fsocie7y: No need for using .order_by("id") here, i think. Quotes will ordered by id by default if u don't change this behaviour

taxi/views.py
```diff
@@ -13,3 +15,30 @@ def index(request):
...
+    paginate_by = 5
+
+
+class CarDetailView(generic.DetailView):
+    model = Car
+
+
+class DriverListView(generic.ListView):
+    model = Driver
+    queryset = Driver.objects.order_by("id")
```
fsocie7y: same as above

fsocie7y (changes requested):

taxi/views.py
nicksetrakov: yeah, ordered by default, but i take this warning without it
photo_2024-04-01_21-11-43

YuriiKindrat (approved):
GJ

PR #574, opened 2024-02-24
Solution

MaxymChyncha:
Manufacturers list
image
Cars list
image
Cars detail
image
Drivers list
image
Drivers detail
image

### DIFF
--- a/taxi/views.py
+++ b/taxi/views.py
@@ -1,4 +1,5 @@
 from django.shortcuts import render
+from django.views import generic
 
 from taxi.models import Driver, Car, Manufacturer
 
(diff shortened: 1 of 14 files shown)


barabarinov (approved):
Great!

taxi/views.py
```diff
@@ -13,3 +14,40 @@ def index(request):
...
 
     return render(request, "taxi/index.html", context=context)
+
+
+class ManufacturerListView(generic.ListView):
+    model = Manufacturer
+    paginate_by = 5
+
+    def get_queryset(self):
+        return Manufacturer.objects.all().order_by("name")
```
Y-Havryliv:
```suggestion
        return Manufacturer.objects.order_by("name")
```

templates/taxi/car_detail.html
```diff
@@ -0,0 +1,16 @@
+{% extends "base.html" %}
+
+{% block content %}
+	<h1>{{ car.model }}</h1>
```
Y-Havryliv: something is wrong with indentation

templates/taxi/car_list.html
```diff
@@ -0,0 +1,16 @@
+{% extends "base.html" %}
+
+{% block content %}
```
Y-Havryliv, line 3: re-read the readme, the task is not done according to it
`Create templates for manufacturer list, car list, driver list. In these templates:
Create table with the information of each instance`

templates/taxi/car_detail.html
Y-Havryliv: check it everywhere

Y-Havryliv (changes requested):

Serhii-Leonenko (approved):

PR #580, opened 2024-02-25
Solution

juliastetsko:
Screenshot 2024-02-26 at 01 04 00
Screenshot 2024-02-26 at 01 03 47
Screenshot 2024-02-26 at 01 03 53
Screenshot 2024-02-26 at 01 03 56

### DIFF
--- a/templates/includes/sidebar.html
+++ b/templates/includes/sidebar.html
@@ -1,6 +1,6 @@
 <ul class="sidebar-nav">
-  <li><a href="#">Home page</a></li>
-  <li><a href="#">Manufacturers</a></li>
-  <li><a href="#">Cars</a></li>
-  <li><a href="#">Drivers</a></li>
+  <li><a href="/">Home page</a></li>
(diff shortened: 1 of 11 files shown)


VladKyianytsia (approved):

templates/includes/sidebar.html
```diff
@@ -1,6 +1,6 @@
 <ul class="sidebar-nav">
-  <li><a href="#">Home page</a></li>
-  <li><a href="#">Manufacturers</a></li>
-  <li><a href="#">Cars</a></li>
-  <li><a href="#">Drivers</a></li>
+  <li><a href="/">Home page</a></li>
```
Y-Havryliv, line 2: sorry, I recalled after the QnA session yesterday, that it's better to use href="{% url 'some_name' %} instead of hard-coded URLs because in case URLs will change in the future, you won't need to change the templates

Y-Havryliv (approved):

PR #615, opened 2024-03-16
Solution

mkuchuman:
image
image
image
image
image
image


mkuchuman:
Could you please answer why I need to use 2 indents at the end of each .html file? Just interesting

vasylhnatiuk (changes requested):
you should use newblank  because of rules of pep

https://www.flake8rules.com/rules/W292.html

mkuchuman:
Thanks for the answer. Should I fix something in the code?
I seem to have indents everywhere.

vasylhnatiuk (approved):

PR #617, opened 2024-03-18
Develop

Moonralex:


taxi/migrations/0004_alter_driver_options.py
```diff
@@ -0,0 +1,17 @@
...
+    dependencies = [
+        ("taxi", "0003_alter_manufacturer_options"),
+    ]
+
+    operations = [
+        migrations.AlterModelOptions(
+            name="driver",
+            options={"ordering": ("first_name",)},
+        ),
+    ]
```
Dimosphen1: It is a good practice to include one migration per app per PR in Django projects since it helps to avoid `migrations` folder overload in the future. Represent all changes done to models in one migration file per app per PR, please

taxi/admin.py
```diff
@@ -27,8 +30,20 @@ class DriverAdmin(UserAdmin):
 
 @admin.register(Car)
 class CarAdmin(admin.ModelAdmin):
+    list_display = ("model", "manufacturer_name",)
     search_fields = ("model",)
     list_filter = ("manufacturer",)
 
+    def get_queryset(self, request: HttpRequest) -> QuerySet:
```
Dimosphen1: Improve your return type annotations to indicate instances of what class you return in the `QuerySet`

Dimosphen1 (changes requested):
Several changes were requested. Besides that, use more meaningful commit messages in the future

taxi/migrations/0004_alter_driver_options.py
Moonralex: Thanks, i didn`t think about that. Now it looks better.

taxi/admin.py
```diff
@@ -1,5 +1,8 @@
 from django.contrib import admin
 from django.contrib.auth.admin import UserAdmin
+from django.db.models import QuerySet
+from django.http import HttpRequest
+
 from .models import Driver, Car, Manufacturer
```
pavlopro: Please use full import paths instead of relative

taxi/urls.py
```diff
@@ -1,9 +1,30 @@
 from django.urls import path
 
-from .views import index
+from .views import (
```
pavlopro: The same about full import here as well

taxi/views.py
```diff
@@ -13,3 +15,33 @@ def index(request):
     }
 
     return render(request, "taxi/index.html", context=context)
+
+
+class ManufacturerListView(generic.ListView):
+    model = Manufacturer
+    template_name = "taxi/manufacturer_list.html"
```
pavlopro: There's no need to define template_name explicitly. It would default to "taxi/manufacturer_list.html"

pavlopro (changes requested):

Dimosphen1 (approved):
LGTM, well done!

PR #624, opened 2024-03-28
Solution

AlexTsikhun:
image
image
image
image
image
image
image

### DIFF
--- a/taxi/urls.py
+++ b/taxi/urls.py
@@ -1,11 +1,13 @@
 from django.urls import path
 
-from .views import (index,
-                    ManufacturerListView,
-                    CarListView,
-                    CarDetailView,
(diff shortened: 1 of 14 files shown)


taxi/urls.py
```diff
@@ -1,9 +1,23 @@
 from django.urls import path
 
-from .views import index
+from .views import (index,
```
pavlopro: Correct the imports so that each one starts on a new line

pavlopro: Use absolute imports instead of relative imports.

pavlopro (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #656, opened 2024-05-03
Develop

vitalinamal:
image
image
image
image
image
image
image


taxi/views.py
```diff
@@ -13,3 +14,33 @@ def index(request):
...
+
+
+class DriverListView(generic.ListView):
+    model = Driver
+    paginate_by = 5
+
+
+class DriverDetailView(generic.DetailView):
+    model = Driver
+    queryset = Driver.objects.prefetch_related()
```
MNDonut: It's better to pass what fields you want to select/prefetch because otherwise Django will fetch all foreign keys and many to many fields. For now it's ok as you have just a few fields, but when you add new foreign keys, they'll be automatically fetched. In case you add what fields you want to fetch, this issue won't happen

MNDonut (changes requested):

taxi/views.py
```diff
@@ -13,3 +14,33 @@ def index(request):
...
+    paginate_by = 5
+
+
+class ManufacturerDetailView(generic.DetailView):
+    model = Manufacturer
+
+
+class CarListView(generic.ListView):
+    model = Car
+    queryset = Car.objects.select_related()
```
MNDonut: select related without fields

MNDonut (changes requested):

MNDonut (approved):

PR #666, opened 2024-05-08
solution v0.90

Chub-andrew:


taxi/urls.py
```diff
@@ -1,9 +1,23 @@
 from django.urls import path
 
 from .views import index
+from taxi.views import (ManufacturerListView,
+                        CarListView,
+                        CarDetailView,
+                        DriverListView,
+                        DriverDetailView)
```
MNDonut: it's recommended to start imported things from the next line

taxi/views.py
```diff
@@ -13,3 +14,33 @@ def index(request):
...
+    paginate_by = 5
+
+
+class CarListView(generic.ListView):
+    model = Car
+    queryset = Car.objects.select_related("manufacturer")
+    paginate_by = 5
+
+    def get_queryset(self):
+        return Car.objects.all().order_by("model")
```
MNDonut: why do you set queryset and override get_queryset?
by default django calls get_queryset that by default will use queryset attribute to get gueryset but in your case your get_queryset doesn't use queryset attribute, thus select related won't be used here

taxi/views.py
```diff
@@ -13,3 +14,33 @@ def index(request):
...
+
+class DriverListView(generic.ListView):
+    model = Driver
+    paginate_by = 5
+    queryset = Driver.objects.all().order_by("username")
+
+
+class DriverDetailView(generic.DetailView):
+    model = Driver
+    queryset = Driver.objects.prefetch_related("cars__manufacturer")
```
MNDonut, line 42: prefetch cars as well?

taxi/views.py
```diff
@@ -13,3 +14,33 @@ def index(request):
...
+
+
+class CarDetailView(generic.DetailView):
+    model = Car
+
+
+class DriverListView(generic.ListView):
+    model = Driver
+    paginate_by = 5
+    queryset = Driver.objects.all().order_by("username")
```
MNDonut: all is redundant. By default when you build a query, django automatically calls .all() + without all it's more optimized because instead of making 2 SQL queries like;
fetch all manufacturers
order them by name
It'll make 1 query: fetch all manufacturers and order them by name at the same time

templates/taxi/driver_list.html
```diff
@@ -0,0 +1,23 @@
...
+      {% if page_obj.has_previous %}
+        <li class="page-item"><a class="page-link" href="?page={{ page_obj.previous_page_number }}">Prev</a></li>
+  {% endif %}
+  <li class="page-item active"><span class="page-link">{{ page_obj.number }}</span></li>
+  {% if page_obj.has_next %}
+    <li class="page-item"><a class="page-link" href="?page={{ page_obj.next_page_number }}">Next</a></li>
+  {% endif %}  
+  </ul>
+  {% endif %}
+{% endblock %}
```
MNDonut: blank line at the end of file

templates/taxi/driver_detail.html
```diff
@@ -0,0 +1,18 @@
...
+    <p><strong>Cars:</strong></p>
+    <ul>
+      {% for car in driver.cars.all %}
+        <li><a href="{% url 'taxi:driver-detail' car.id %}">{{ car.model }}</a></li>
+      {% endfor %}
+    </ul>
+  {% else %}
+    <p>No drivers</p>
+  {% endif %}
+{% endblock %}
```
MNDonut: blank line at the end of file

templates/taxi/car_list.html
```diff
@@ -0,0 +1,23 @@
...
+      {% if page_obj.has_previous %}
+        <li class="page-item"><a class="page-link" href="?page={{ page_obj.previous_page_number }}">Prev</a></li>
+  {% endif %}
+  <li class="page-item active"><span class="page-link">{{ page_obj.number }}</span></li>
+  {% if page_obj.has_next %}
+    <li class="page-item"><a class="page-link" href="?page={{ page_obj.next_page_number }}">Next</a></li>
+  {% endif %}  
+  </ul>
+  {% endif %}
+{% endblock %}
```
MNDonut: blank line at the end of file

templates/taxi/car_detail.html
```diff
@@ -0,0 +1,18 @@
...
+    <p><strong>Drivers:</strong></p>
+    <ul>
+      {% for driver in car.drivers.all %}
+        <li><a href="{% url 'taxi:car-detail' driver.id %}"> {{ driver.first_name }} {{ driver.last_name }}</a></li>
+      {% endfor %}
+    </ul>
+  {% else %}
+    <p>No drivers</p>
+  {% endif %}
+{% endblock %}
```
MNDonut: blank line at the end of file

MNDonut (changes requested):

taxi/views.py
Chub-andrew: I created this method so that during testing, not appear Warnings: about order_by

YuriiKindrat: u could use order_by method directly in the queryset attr, no need to overwrite method here

YuriiKindrat (changes requested):

YuriiKindrat (approved):
GJ

PR #671, opened 2024-05-09
Solution

siaserb:


taxi/urls.py
```diff
@@ -1,9 +1,14 @@
 from django.urls import path
 
-from .views import index
+from .views import index, CarListView, CarDetailView, ManufacturerListView, DriverListView, DriverDetailView
```
MNDonut: I suppose it's longer than 79 characters(recommendation by PEP8), thus you have 2 choices:
1. move each view on a separate line
2. from . import views, and then for each url use view.VIEW_NAME

taxi/urls.py
```diff
@@ -1,9 +1,14 @@
...
 
 urlpatterns = [
     path("", index, name="index"),
+    path("manufacturer/", ManufacturerListView.as_view(), name="manufacturer-list"),
+    path("cars/", CarListView.as_view(), name="cars-list"),
+    path("cars/<int:pk>", CarDetailView.as_view(), name="cars-detail"),
+    path("drivers/", DriverListView.as_view(), name="drivers-list"),
+    path("drivers/<int:pk>", DriverDetailView.as_view(), name="drivers-detail"),
```
MNDonut: don't forget to add a slash at the end of each url

taxi/views.py
```diff
@@ -13,3 +14,29 @@ def index(request):
     }
 
     return render(request, "taxi/index.html", context=context)
+
+
+class ManufacturerListView(ListView):
+    model = Manufacturer
+    queryset = Manufacturer.objects.all().order_by("name")
```
MNDonut: all is redundant. By default when you build a query, django automatically calls .all() + without all it's more optimized because instead of making 2 SQL queries like:
fetch all manufacturers
order them by name
It'll make 1 query: fetch all manufacturers and order them by name at the same time

taxi/views.py
```diff
@@ -13,3 +14,29 @@ def index(request):
...
+
+
+class DriverListView(ListView):
+    model = Driver
+    paginate_by = 5
+
+
+class DriverDetailView(DetailView):
+    model = Driver
+    queryset = Driver.objects.prefetch_related("cars__manufacturer")
```
MNDonut: prefetch cars as well?

MNDonut (changes requested):

MNDonut (changes requested):
And tests failed :)

YuriiKindrat (approved):
GJ

PR #714, opened 2024-06-29
Solution

Viesich:


taxi/urls.py
```diff
@@ -1,9 +1,29 @@
 from django.urls import path
+from taxi.views import (
```
viktoria-rybenchuk, line 3: add a blank line between different group of imports

taxi/views.py
```diff
@@ -13,3 +15,33 @@ def index(request):
     }
 
     return render(request, "taxi/index.html", context=context)
+
+
+class ManufacturerListView(generic.ListView):
+    model = Manufacturer
+    queryset = Manufacturer.objects.all().order_by("name",)
```
viktoria-rybenchuk: `all()` is redundant

taxi/views.py
```diff
@@ -13,3 +15,33 @@ def index(request):
...
+    paginate_by = 5
+
+
+class CarListView(generic.ListView):
+    model = Car
+    queryset = Car.objects.select_related("manufacturer").all().distinct()
```
viktoria-rybenchuk: all() is redundant

taxi/views.py
```diff
@@ -13,3 +15,33 @@ def index(request):
...
+    paginate_by = 5
+
+
+class CarDetailView(generic.DetailView):
+    model = Car
+    template_name = "taxi/car_detail.html"
+    context_object_name = "car"
```
viktoria-rybenchuk: No need to redefine this variables, django handles it by default

taxi/views.py
```diff
@@ -13,3 +15,33 @@ def index(request):
...
+class DriverListView(generic.ListView):
+    model = Driver
+    paginate_by = 5
+
+
+class DriverDetailView(generic.DetailView):
+    model = Driver
+    template_name = "taxi/driver_detail.html"
+    context_object_name = "driver"
+    queryset = Driver.objects.all()
```
viktoria-rybenchuk: the same here

viktoria-rybenchuk (changes requested):

taxi/views.py
Viesich: I understand, thanks!)
This is all implemented "under the hood" in generic

taxi/views.py
Viesich: all() is not needed here, because select_related returns data from the join table to the queryset)

viktoria-rybenchuk (approved):
Great!

PR #717, opened 2024-07-01
py-ts-class-based-views v1

BornToLivee:


BornToLivee:
image

BornToLivee:
image

BornToLivee:
image

BornToLivee:
image

BornToLivee:
image

taxi/views.py
```diff
@@ -13,3 +15,29 @@ def index(request):
     }
 
     return render(request, "taxi/index.html", context=context)
+
+
+class ManufacturerListView(ListView):
+    model = Manufacturer
+    queryset = Manufacturer.objects.all().order_by("name")
```
Arsen-hrynevych, line 22: In Django, using `.all()` before `.order_by()` is unnecessary. The `.order_by()` method already includes all the results by default. Therefore, you can simplify `Manufacturers.objects.all().order_by('name')` to `Manufacturers.objects.order_by('name')

taxi/urls.py
```diff
@@ -1,9 +1,16 @@
 from django.urls import path
 
-from .views import index
+from .views import (index, ManufacturerListView, CarListView,
```
Arsen-hrynevych, line 3: If importing more than two items, it is better to write them on separate lines. This improves readability and maintenance of the code. 

Example:

From taxi.models import ( 
	Driver,
	Manufacturer,
	Car
)

Arsen-hrynevych (approved):
Good job! Just fix minor style issues. Once you make this change, it will be perfect.

PR #719, opened 2024-07-07
Solution

BohdanKuzik:


DevSpaciX (changes requested):
Attach pls screenshots of your app )
image

taxi/views.py
```diff
@@ -13,3 +15,29 @@ def index(request):
     }
 
     return render(request, "taxi/index.html", context=context)
+
+
+class ManufacturerListView(generic.ListView):
+    model = Manufacturer
+    queryset = Manufacturer.objects.all().order_by("name")
+    paginate_by = 5
+
```
DevSpaciX: .all() is not necessary here... If .order_by() is. Otherwise it's the latter that should be removed.

DevSpaciX (changes requested):

templates/taxi/driver_list.html
```diff
@@ -0,0 +1,16 @@
+{% extends "base.html" %}
+{% block content %}
+  <h1>Driver list</h1>
+  {% if driver_list %}
+    <ul>
+      {% for driver in driver_list%}
+        <li>
+          <a href="{% url "taxi:driver-detail" pk=driver.id %}">{{ driver.username }}</a>
```
Arsen-hrynevych, line 8: When you use double quotes "" inside of double quotes, you can only use single quotes ''.

Arsen-hrynevych (approved):
Good job! I really appreciate your effort and attention to detail. Thank you for your hard work. Please fix minor issue

PR #723, opened 2024-07-08
solution

NikitaNikolaitsev:


NikitaNikolaitsev:
Screenshot_40

taxi/urls.py
```diff
@@ -1,9 +1,17 @@
 from django.urls import path
 
-from .views import index
+from .views import (index, ManufacturerView, CarListView,
+                    CarDetailView, DriverListView, DriverDetailView)
```

Example:

From taxi.models import ( 
	Driver,
	Manufacturer,
	Car
)

taxi/views.py
```diff
@@ -13,3 +15,29 @@ def index(request):
     }
 
     return render(request, "taxi/index.html", context=context)
+
+
+class ManufacturerView(generic.ListView):
+    model = Manufacturer
+    queryset = Manufacturer.objects.all().order_by("name")
```
Arsen-hrynevych: When you use `order_by`, it by default returns all objects that satisfy the rule. Therefore, you don't need to explicitly use `.all()`.

taxi/views.py
```diff
@@ -13,3 +15,29 @@ def index(request):
...
+    paginate_by = 5
+
+
+class CarListView(generic.ListView):
+    model = Car
+    paginate_by = 5
+    queryset = Car.objects.select_related()
```
Arsen-hrynevych: When select_related is empty it doesn't do anything.

taxi/views.py
```diff
@@ -13,3 +15,29 @@ def index(request):
...
+
+
+class DriverListView(generic.ListView):
+    model = Driver
+    paginate_by = 5
+
+
+class DriverDetailView(generic.DetailView):
+    model = Driver
+    queryset = Driver.objects.select_related()
```
Arsen-hrynevych: When select_related is empty it doesn't do anything.

templates/taxi/car_list.html
```diff
@@ -0,0 +1,20 @@
...
+  <h1>Cars</h1>
+  {% if car_list %}
+    <table>
+      <tr>
+        <th>id</th>
+        <th>Model</th>
+      </tr>
+    {% for car in car_list %}
+      <tr>
+        <th><a href="{% url "taxi:car-detail" car.id %}">{{ car.id }}</a></th>
```

Example:

Check correctnes of using quoutes because you use "" and inside "" again

Arsen-hrynevych (changes requested):
Good job! I really appreciate your effort . Please review the check list of the task and comments.

taxi/urls.py
```diff
@@ -1,9 +1,19 @@
 from django.urls import path
 
-from .views import index
+from .views import (
+    index, ManufacturerView, CarListView,
+    CarDetailView, DriverListView, DriverDetailView
+)
```
Arsen-hrynevych, line 6: If importing more than two items, it is better to write them on separate lines. This improves readability and maintenance of the code. 

Example:

From taxi.models import ( 
	Driver,
	Manufacturer,
	Car
)

templates/includes/sidebar.html
```diff
@@ -1,6 +1,6 @@
 <ul class="sidebar-nav">
-  <li><a href="#">Home page</a></li>
-  <li><a href="#">Manufacturers</a></li>
-  <li><a href="#">Cars</a></li>
-  <li><a href="#">Drivers</a></li>
+  <li><a href="{% url "taxi:index" %}">Home page</a></li>
+  <li><a href="{% url "taxi:manufacturer-list" %}">Manufacturers</a></li>
+  <li><a href="{% url "taxi:car-list" %}">Cars</a></li>
+  <li><a href="{% url "taxi:driver-list" %}">Drivers</a></li>
```
Arsen-hrynevych, line 5: Check correctnes of using quoutes because you use "" and inside "" again

Arsen-hrynevych (approved):

PR #725, opened 2024-07-08
add solution

jpdreamthug:
Manufacturer list, order by name
image
Cars list:
image
Car detail:
image
Drivers list:
image
Driver detail:
image

### DIFF
--- a/taxi/urls.py
+++ b/taxi/urls.py
@@ -1,9 +1,29 @@
 from django.urls import path
 
-from .views import index
+from .views import (index,
+                    ManufacturerListView,
+                    CarListView,
+                    DriverListView,
+                    CarDetailView,
+                    DriverDetailView)
 
 urlpatterns = [
     path("", index, name="index"),
+    path("manufacturers/",
+         ManufacturerListView.as_view(),
+         name="manufacturer-list"),
+    path("cars/",
+         CarListView.as_view(),
+         name="car-list"),
+    path("cars/<int:pk>/",
+         CarDetailView.as_view(),
+         name="car-detail"),
+    path("drivers/",
+         DriverListView.as_view(),
+         name="driver-list"),
+    path("drivers/<int:pk>/",
+         DriverDetailView.as_view(),
+         name="driver-detail"),
 ]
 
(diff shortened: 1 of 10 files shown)


taxi/urls.py
```diff
@@ -1,9 +1,29 @@
 from django.urls import path
 
-from .views import index
+from .views import (index,
+                    ManufacturerListView,
+                    CarListView,
+                    DriverListView,
+                    CarDetailView,
+                    DriverDetailView)
```

Example:

From taxi.models import ( 
	Driver,
	Manufacturer,
	Car
)

templates/taxi/car_list.html
```diff
@@ -0,0 +1,11 @@
+{% extends "base.html" %}
+{% block content %}
+  <h1>Cars List</h1>
+  <ul>
+  {% for car in car_list %}
+    <li><a href="{% url 'taxi:car-detail' car.id %}">{{ car.model }}</a></li>
```

Example:

templates/taxi/driver_list.html
```diff
@@ -0,0 +1,11 @@
+{% extends "base.html" %}
+{% block content %}
+  <h1>Drivers List</h1>
+  <ul>
+  {% for driver in driver_list %}
+    <li><a href="{% url 'taxi:driver-detail' driver.id %}">{{ driver.username }}</a></li>
```
Arsen-hrynevych: The {% url %} template tag generates URLs dynamically using the URL name defined in your urls.py file. The parameters you pass must match the named parameters expected by the URL pattern. 

Example:

Arsen-hrynevych (changes requested):
Good job! I really appreciate your effort . Please review the comments and fix minor issues.

taxi/urls.py
jpdreamthug: Hi, I looked at this link, can you tell me more specifically where I should look for information about this?

templates/taxi/car_list.html
jpdreamthug: Will this have any effect on the code? I thought it was the same thing.

Arsen-hrynevych (approved):

PR #726, opened 2024-07-08
Solution_v1

BileichukIvan:


BileichukIvan:
image
image
image
image
image
image

taxi/views.py
```diff
@@ -13,3 +15,29 @@ def index(request):
...
+
+
+class DriverListView(generic.ListView):
+    model = Driver
+    paginate_by = 5
+
+
+class DriverDetailView(generic.DetailView):
+    model = Driver
+    queryset = Driver.objects.select_related()
```

taxi/views.py
```diff
@@ -13,3 +15,29 @@ def index(request):
     }
 
     return render(request, "taxi/index.html", context=context)
+
+
+class ManufacturerListView(generic.ListView):
+    model = Manufacturer
+    queryset = Manufacturer.objects.all().order_by("name")
```

taxi/urls.py
```diff
@@ -1,9 +1,23 @@
 from django.urls import path
 
-from .views import index
+from .views import (index, ManufacturerListView, CarListView,
+                    CarDetailView, DriverListView, DriverDetailView)
```

Example:

From taxi.models import ( 
	Driver,
	Manufacturer,
	Car
)

Arsen-hrynevych (changes requested):
Good job! I really appreciate your effort . Please review the check list of the task and fix minor issues.

DevSpaciX (approved):

PR #727, opened 2024-07-08
Solution

VShveda:
image
image
image
image

### DIFF
--- a/taxi/urls.py
+++ b/taxi/urls.py
@@ -1,8 +1,13 @@
 from django.urls import path
 
-from .views import (index, ManufacturerListView,
-                    CarListView, CarDetailView,
-                    DriverListView, DriverDetailView)
+from .views import (
+    index,
+    ManufacturerListView,
+    CarListView,
+    CarDetailView,
+    DriverListView,
+    DriverDetailView
+)
 
 urlpatterns = [
     path("", index, name="index"),
--- a/taxi/urls.py
+++ b/taxi/urls.py
@@ -1,9 +1,24 @@
 from django.urls import path
 
-from .views import index
+from .views import (index, ManufacturerListView,
+                    CarListView, CarDetailView,
+                    DriverListView, DriverDetailView)
 
 urlpatterns = [
     path("", index, name="index"),
+    path(
+        "manufacturers/",
+        ManufacturerListView.as_view(),
+        name="manufacturer-list"
+    ),
+    path("cars/", CarListView.as_view(), name="car-list"),
+    path("cars/<int:pk>/", CarDetailView.as_view(), name="car-detail"),
+    path("drivers/", DriverListView.as_view(), name="driver-list"),
+    path(
+        "drivers/<int:pk>/",
+        DriverDetailView.as_view(),
+        name="driver-detail"
+    ),
 ]
 
 app_name = "taxi"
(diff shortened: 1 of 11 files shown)


taxi/urls.py
```diff
@@ -1,9 +1,24 @@
 from django.urls import path
 
-from .views import index
+from .views import (index, ManufacturerListView,
+                    CarListView, CarDetailView,
+                    DriverListView, DriverDetailView)
```

Example:

From taxi.models import ( 
	Driver,
	Manufacturer,
	Car
)

taxi/views.py
```diff
@@ -13,3 +15,29 @@ def index(request):
     }
 
     return render(request, "taxi/index.html", context=context)
+
+
+class ManufacturerListView(generic.ListView):
+    model = Manufacturer
+    queryset = Manufacturer.objects.all().order_by("name")
```

taxi/views.py
```diff
@@ -13,3 +15,29 @@ def index(request):
...
+
+
+class DriverListView(generic.ListView):
+    model = Driver
+    paginate_by = 5
+
+
+class DriverDetailView(generic.DetailView):
+    model = Driver
+    queryset = Driver.objects.select_related()
```

Arsen-hrynevych (changes requested):
Good job! I really appreciate your effort . Please review the check list of the task and fix minor issues.

taxi/views.py
```diff
@@ -1,9 +1,10 @@
 from django.shortcuts import render
-
+from django.http import HttpResponse, HttpRequest
+from django.views import generic
 from taxi.models import Driver, Car, Manufacturer
```
Arsen-hrynevych: Imports should be grouped and separated by a blank line according to standard library, external libraries, and custom imports. This practice enhances both code readability and maintainability. 

taxi/views.py
```diff
@@ -13,3 +14,29 @@ def index(request):
...
+
+
+class DriverListView(generic.ListView):
+    model = Driver
+    paginate_by = 5
+
+
+class DriverDetailView(generic.DetailView):
+    model = Driver
+    queryset = Driver.objects.all()
```
Arsen-hrynevych: Please review the checklist once more. In this view, you display information about cars of the driver. Optimize query: don't make a query for a manufacturer for each car (N+1 problem).

templates/includes/sidebar.html
```diff
@@ -1,6 +1,6 @@
 <ul class="sidebar-nav">
-  <li><a href="#">Home page</a></li>
-  <li><a href="#">Manufacturers</a></li>
-  <li><a href="#">Cars</a></li>
-  <li><a href="#">Drivers</a></li>
+  <li><a href={% url "taxi:index" %}>Home page</a></li>
+  <li><a href={% url "taxi:manufacturer-list" %}>Manufacturers</a></li>
+  <li><a href={% url "taxi:car-list" %}>Cars</a></li>
+  <li><a href={% url "taxi:driver-list" %}>Drivers</a></li>
```
Arsen-hrynevych: Check correctnes of using quoutes because you use "" and inside "" again

Arsen-hrynevych (changes requested):

templates/includes/sidebar.html
VShveda: I don't understand where I used "" inside ""

VShveda: Never mind, I get it.

Arsen-hrynevych: Sorry that was my mistake :) But the mistake still exist https://stackoverflow.com/questions/17200389/how-to-use-the-href-attribute-in-django-templates

Arsen-hrynevych (approved):

PR #730, opened 2024-07-09
Solution

b3v3kt0r:


b3v3kt0r:
image
image
image
image

taxi/urls.py
```diff
@@ -1,9 +1,24 @@
 from django.urls import path
 
-from .views import index
+from .views import (index,
+                    ManufacturerListView,
+                    CarListView,
+                    CarDetailView,
+                    DriverListView,
+                    DriverDetailView)
```

Example:

From taxi.models import ( 
	Driver,
	Manufacturer,
	Car
)

taxi/views.py
```diff
@@ -13,3 +14,29 @@ def index(request):
     }
 
     return render(request, "taxi/index.html", context=context)
+
+
+class ManufacturerListView(generic.ListView):
+    model = Manufacturer
+    queryset = Manufacturer.objects.all().order_by("name")
```

taxi/views.py
```diff
@@ -13,3 +14,29 @@ def index(request):
...
+
+
+class DriverListView(generic.ListView):
+    model = Driver
+    paginate_by = 5
+
+
+class DriverDetailView(generic.DetailView):
+    model = Driver
+    queryset = Driver.objects.select_related()
```

Arsen-hrynevych (changes requested):
Good job! I really appreciate your effort and attention to detail. Thank you for your hard work. Fix the minor issues and that's it.

templates/includes/sidebar.html
```diff
@@ -1,6 +1,6 @@
 <ul class="sidebar-nav">
-  <li><a href="#">Home page</a></li>
-  <li><a href="#">Manufacturers</a></li>
-  <li><a href="#">Cars</a></li>
-  <li><a href="#">Drivers</a></li>
+  <li><a href={% url "taxi:index" %}>Home page</a></li>
+  <li><a href={% url "taxi:manufacturer-list" %}>Manufacturers</a></li>
+  <li><a href={% url "taxi:car-list" %}>Cars</a></li>
+  <li><a href={% url "taxi:driver-list" %}>Drivers</a></li>
```
Arsen-hrynevych, line 5: Check correctnes of using quoutes because you use "" and inside "" again

Arsen-hrynevych (approved):
Good job! I really appreciate your effort . Please review fix quoutes.

PR #733, opened 2024-07-09
solution

Andrii1000:


Andrii1000:
image
image
image
image
image
image

taxi/urls.py
```diff
@@ -1,9 +1,23 @@
 from django.urls import path
 
-from .views import index
+from .views import (index,
+                    ManufacturerListView,
+                    CarListView,
+                    CarDetailView,
+                    DriverListView,
+                    DriverDetailView)
```

Example:

From taxi.models import ( 
	Driver,
	Manufacturer,
	Car
)

templates/taxi/car_list.html
```diff
@@ -0,0 +1,13 @@
+{% extends "base.html" %}
+{% block content %}
+  <h1>Car list</h1>
+  {% if car_list %}
+    <ul>
+      {% for car in car_list %}
+        <li><a href="{{ car.id }}">{{ car.id }}</a> </li>
```
Arsen-hrynevych: I noticed a small mistake in your code. In the task, you were supposed to use a URL template tag to generate the link for the car detail page. In your code, the href attribute is set directly to car.id, which won't create the correct link to the car detail page. The url template tag is needed to generate the appropriate URL.

templates/taxi/driver_list.html
```diff
@@ -0,0 +1,13 @@
+{% extends "base.html" %}
+{% block content %}
+  <h1>Drivers list</h1>
+  {% if driver_list %}
+    <ul>
+      {% for driver in driver_list %}
+        <li><a href="{{ driver.id }}">{{ driver.username }}</a></li>
```
Arsen-hrynevych: I noticed a small mistake in your code. In the task, you were supposed to use a URL template tag to generate the link for the driver detail page. In your code, the href attribute is set directly to car.id, which won't create the correct link to the car detail page. The url template tag is needed to generate the appropriate URL.

Arsen-hrynevych (changes requested):

templates/taxi/car_list.html
Andrii1000: I changed it if I correct understand

Arsen-hrynevych (approved):

PR #736, opened 2024-07-09
"Solution"

MaglorFeanorson:
dz2222
dz22222
dz2
dz22
dz222

### DIFF
--- a/taxi/urls.py
+++ b/taxi/urls.py
@@ -1,7 +1,13 @@
 from django.urls import path
 
-from .views import (index, ManufacturerListView, CarListView,
-                    CarDetailView, DriverListView, DriverDetailView)
+from .views import (
+    index,
+    ManufacturerListView,
+    CarListView,
+    CarDetailView,
+    DriverListView,
+    DriverDetailView
+)
 
 urlpatterns = [
     path("", index, name="index"),
--- a/taxi/urls.py
+++ b/taxi/urls.py
@@ -1,9 +1,22 @@
 from django.urls import path
 
-from .views import index
+from .views import (index, ManufacturerListView, CarListView,
+                    CarDetailView, DriverListView, DriverDetailView)
 
 urlpatterns = [
     path("", index, name="index"),
+    path(
+        "manufacturers/",
+        ManufacturerListView.as_view(),
+        name="manufacturer-list"
+    ),
+    path("cars/", CarListView.as_view(), name="car-list"),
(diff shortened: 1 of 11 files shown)


taxi/urls.py
```diff
@@ -1,9 +1,22 @@
 from django.urls import path
 
-from .views import index
+from .views import (index, ManufacturerListView, CarListView,
+                    CarDetailView, DriverListView, DriverDetailView)
```

Example:

From taxi.models import ( 
	Driver,
	Manufacturer,
	Car
)

taxi/views.py
```diff
@@ -1,15 +1,39 @@
...
+
+
+class DriverListView(generic.ListView):
+    model = Driver
+    paginate_by = 5
+
+
+class DriverDetailView(generic.DetailView):
+    model = Driver
+    queryset = Driver.objects.select_related()
```
Arsen-hrynevych: An empty `select_related` doesn't do anything. Make sure to double-check if you are using `select_related` correctly.

Read more [here](https://medium.com/@soverignchriss/understanding-select-related-and-prefetch-related-methods-in-django-orm-db36003d5fbf).

templates/taxi/car_list.html
```diff
@@ -0,0 +1,22 @@
...
+  {% if car_list %}
+    <table>
+      <tr>
+        <th>id</th>
+        <th>model</th>
+        <th>manufacturer</th>
+      </tr>
+      {% for car in car_list %}
+        <tr>
+          <th><a href="{{ car.get_absolute_url }}">{{ car.id }}</a></th>
```
Arsen-hrynevych: I noticed a mistake in your code regarding the URL generation for the car detail page. Instead of using the get_absolute_url method, you should use the url template tag with the correct parameters.

templates/taxi/driver_list.html
```diff
@@ -0,0 +1,24 @@
...
+      <tr>
+        <th>id</th>
+        <th>username</th>
+        <th>first name</th>
+        <th>last name</th>
+      </tr>
+      {% for driver in driver_list %}
+        <tr>
+          <th>{{ driver.id }}</th>
+          <th><a href="{{ driver.get_absolute_url }}">{{ driver.username }}</a></th>
```
Arsen-hrynevych: Please review the checklist of the task

Arsen-hrynevych (changes requested):

DevSpaciX (approved):

PR #737, opened 2024-07-09
Solution v.1

SemenetsSergii:
image
image
image
image
image
image
image
image

### DIFF
--- a/taxi/urls.py
+++ b/taxi/urls.py
@@ -1,13 +1,13 @@
 from django.urls import path
 
 from .views import (
-                    index,
-                    ManufacturerListView,
-                    CarListView,
-                    CarDetailView,
-                    DriverListView,
-                    DriverDetailView
-                    )
+    index,
+    ManufacturerListView,
+    CarListView,
+    CarDetailView,
+    DriverListView,
+    DriverDetailView
+)
 
 urlpatterns = [
     path("", index, name="index"),
--- a/taxi/urls.py
+++ b/taxi/urls.py
@@ -1,9 +1,31 @@
 from django.urls import path
 
-from .views import index
+from .views import (index,
+                    ManufacturerListView,
+                    CarListView,
+                    CarDetailView,
+                    DriverListView,
+                    DriverDetailView
+                    )
 
 urlpatterns = [
     path("", index, name="index"),
+    path("manufacturers/",
+         ManufacturerListView.as_view(),
+         name="manufacturer-list"),
+    path("cars/",
+         CarListView.as_view(),
+         name="car-list"),
+    path("cars/<int:pk>/",
+         CarDetailView.as_view(),
+         name="car-detail"),
+    path("drivers/",
+         DriverListView.as_view(),
+         name="driver-list"),
+    path("drivers/<int:pk>/",
+         DriverDetailView.as_view(),
+         name="driver-detail"),
 ]
+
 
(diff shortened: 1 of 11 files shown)


taxi/views.py
```diff
@@ -13,3 +15,29 @@ def index(request):
     }
 
     return render(request, "taxi/index.html", context=context)
+
+
+class ManufacturerListView(generic.ListView):
+    model = Manufacturer
+    queryset = Manufacturer.objects.all().order_by("name")
```

taxi/views.py
```diff
@@ -13,3 +15,29 @@ def index(request):
...
+    paginate_by = 5
+
+
+class CarListView(generic.ListView):
+    model = Car
+    queryset = Car.objects.select_related()
```

taxi/views.py
```diff
@@ -13,3 +15,29 @@ def index(request):
...
+
+
+class DriverListView(generic.ListView):
+    model = Driver
+    paginate_by = 5
+
+
+class DriverDetailView(generic.DetailView):
+    model = Driver
+    queryset = Driver.objects.select_related()
```
Arsen-hrynevych: An empty `select_related` doesn't do anything. Make sure to double-check if you are using `select_related` correctly.

Read more [here](https://medium.com/@soverignchriss/understanding-select-related-and-prefetch-related-methods-in-django-orm-db36003d5fbf).

templates/includes/sidebar.html
```diff
@@ -1,6 +1,6 @@
 <ul class="sidebar-nav">
-  <li><a href="#">Home page</a></li>
-  <li><a href="#">Manufacturers</a></li>
-  <li><a href="#">Cars</a></li>
-  <li><a href="#">Drivers</a></li>
+  <li><a href={% url "taxi:index" %}>Home page</a></li>
+  <li><a href={% url "taxi:manufacturer-list" %}>Manufacturers</a></li>
+  <li><a href={% url "taxi:car-list" %}>Cars</a></li>
+  <li><a href={% url "taxi:driver-list" %}>Drivers</a></li>
```
Arsen-hrynevych: Check correctnes of using quoutes because you use "" and inside "" again

templates/taxi/car_detail.html
```diff
@@ -0,0 +1,23 @@
...
+    <li>ID: {{ car.manufacturer.id }}</li>
+    <li>Name: {{ car.manufacturer.name }}</li>
+    <li>Country: {{ car.manufacturer.country }}</li>
+  </ul>
+  <h2>Drivers:</h2>
+  <ul>
+    {% for driver in car.drivers.all %}
+      <li>
+        <p>ID: {{ driver.id }}</p>
+        <p>Username: {{ driver.username }}</p>
```
Arsen-hrynevych, line 10: Please review the checklist again

taxi/urls.py
```diff
@@ -1,9 +1,31 @@
 from django.urls import path
 
-from .views import index
+from .views import (index,
+                    ManufacturerListView,
+                    CarListView,
+                    CarDetailView,
+                    DriverListView,
+                    DriverDetailView
+                    )
```

Example:

From taxi.models import ( 
	Driver,
	Manufacturer,
	Car
)

Arsen-hrynevych (changes requested):

DevSpaciX (approved):

PR #739, opened 2024-07-11
sol

Oleh323:


Oleh323:
Screenshot_3
Screenshot_4
Screenshot_5
Screenshot_6

taxi/views.py
```diff
@@ -13,3 +15,29 @@ def index(request):
...
+
+
+class DriverListView(generic.ListView):
+    model = Driver
+    paginate_by = 5
+
+
+class DriverDetailView(generic.DetailView):
+    model = Driver
+    queryset = Driver.objects.select_related()
```
DevSpaciX, line 43: Actually it should be perfetch_related() here

taxi/views.py
```diff
@@ -13,3 +15,29 @@ def index(request):
     }
 
     return render(request, "taxi/index.html", context=context)
+
+
+class ManufacturerListView(generic.ListView):
+    model = Manufacturer
+    queryset = Manufacturer.objects.all().order_by("name")
```
DevSpaciX, line 22: Actually you do not need here .all()
It can be just 
`queryset = Manufacturer.objects.order_by("name")
`

DevSpaciX (changes requested):

PR #746, opened 2024-07-18
Develop

pikaOpika:
image
image
image
image
image
image


templates/includes/sidebar.html
```diff
@@ -1,6 +1,6 @@
 <ul class="sidebar-nav">
-  <li><a href="#">Home page</a></li>
-  <li><a href="#">Manufacturers</a></li>
-  <li><a href="#">Cars</a></li>
-  <li><a href="#">Drivers</a></li>
+  <li><a href="/">Home page</a></li>
+  <li><a href="/manufacturers">Manufacturers</a></li>
```
pavlopro: Use {% url %} template tag instead

pavlopro (changes requested):
Add visible styles (e.g. change the background color) here and update the screenshot to indicate that static files are served correctly

pikaOpika:
> Add visible styles (e.g. change the background color) here and update the screenshot to indicate that static files are served correctly

I tried to add styles, but for some reason, they are not working even though I have connected them. Most likely, I missed something. :(

vasylhnatiuk (approved):

PR #759, opened 2024-08-04
Solution

taras-tkachuk:
image
image
image
image
image


taxi/models.py
```diff
@@ -1,19 +1,35 @@
...
+
 
 class Driver(AbstractUser):
     license_number = models.CharField(max_length=255, unique=True)
 
+    class Meta:
+        ordering = ("username", )
+
+    def get_absolute_url(self) -> object:
+        return reverse("taxi:driver-detail", args=[str(self.id)])
```
vsmutok:
```suggestion
        return reverse("taxi:driver-detail",  kwargs={"pk": self.pk})
```

taxi/models.py
```diff
@@ -1,19 +1,35 @@
...
     manufacturer = models.ForeignKey(
         Manufacturer, on_delete=models.CASCADE, related_name="cars"
     )
     drivers = models.ManyToManyField(Driver, related_name="cars")
+
+    class Meta:
+        ordering = ("model", )
+
+    def get_absolute_url(self) -> object:
+        return reverse("taxi:car-detail", args=[str(self.id)])
```
vsmutok:
```suggestion
        return reverse("taxi:car-detail", kwargs={"pk": self.pk})
```
in this case `kwargs` would be better

vsmutok (changes requested):

vasylhnatiuk (approved):

PR #761, opened 2024-08-05
created class-based view for Taxi service

MasakDirt:
### **_Home page ->_** 

image

### **_Manufacturers list page ->_**
 
image

### **_Cars list page ->_**

image

### **_Drivers list page ->_** 

image

### **_Driver detail page ->_**

image

### **_Car detail page ->_**

image

### DIFF
--- a/taxi/urls.py
+++ b/taxi/urls.py
@@ -1,9 +1,23 @@
 from django.urls import path
 
-from .views import index
+from .views import index, ManufacturerListView, CarListView, CarDetailView, \
+    DriverDetailView, DriverListView
 
 urlpatterns = [
     path("", index, name="index"),
+    path("cars/", CarListView.as_view(), name="car-list"),
+    path("cars/<int:pk>/", CarDetailView.as_view(), name="car-detail"),
+    path("drivers/", DriverListView.as_view(), name="driver-list"),
+    path(
+        "manufacturers/",
+        ManufacturerListView.as_view(),
+        name="manufacturer-list"
+    ),
+    path(
+        "drivers/<int:pk>/",
+        DriverDetailView.as_view(),
+        name="driver-detail"
+    ),
 ]
 
 app_name = "taxi"
--- a/taxi/urls.py
+++ b/taxi/urls.py
@@ -1,7 +1,14 @@
 from django.urls import path
 
-from .views import index, ManufacturerListView, CarListView, CarDetailView, \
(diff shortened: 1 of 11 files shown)


taxi/urls.py
```diff
@@ -1,9 +1,23 @@
 from django.urls import path
 
-from .views import index
+from .views import index, ManufacturerListView, CarListView, CarDetailView, \
+    DriverDetailView, DriverListView
```
MNDonut: we don't use \ as a line breaker. use ( ) instead

taxi/urls.py
```diff
@@ -1,9 +1,23 @@
...
+    path(
+        "manufacturers/",
+        ManufacturerListView.as_view(),
+        name="manufacturer-list"
+    ),
+    path(
+        "drivers/<int:pk>/",
+        DriverDetailView.as_view(),
+        name="driver-detail"
+    ),
```
MNDonut: let's move drivers/id higher

taxi/views.py
```diff
@@ -13,3 +14,35 @@ def index(request):
...
+
+class DriverListView(generic.ListView):
+    model = Driver
+    paginate_by = 5
+    template_name = "taxi/driver_list.html"
+
+
+class DriverDetailView(generic.DetailView):
+    model = Driver
+    queryset = Driver.objects.prefetch_related("cars")
```
MNDonut: you prefetch cars, but not manufacturer

MNDonut (changes requested):

taxi/urls.py
MasakDirt: As I understand it is the best way to place similar paths near each other?

MNDonut: @MasakDirt  Yes, if possible. It depends on what urls you have.
For example if you have
`drivers/`
`drivers/<str:name>/`
`drivers/only_active/`
then it's better to move 2 url after the 3 because Django looks for the first pattern that satisfies entered url.
For example user enters this url
`drivers/only_active/`
You have this url, but the first pattern that works is 
`drivers/<str:name>/` (because str:name means a dynamic string, thus Django will use this url and an appropriate view)

So:
1. The same "main" route
2. Static urls
3. Dynamic urls

And the correct ordering should be
`drivers/`
`drivers/only_active/`
`drivers/<str:name>/`

MNDonut (approved):
Left a comment regarding urls ordering

PR #773, opened 2024-08-14
Solution

oboliziuk:
cars
drivers
manufacturers

### DIFF
--- a/taxi/urls.py
+++ b/taxi/urls.py
@@ -1,15 +1,20 @@
 from django.urls import path
 
-from .views import (index, CarListView,
-                    ManufacturerListView, car_detail_view,
-                    DriverListView, DriverDetailView)
+from .views import (
+    index,
+    CarListView,
+    ManufacturerListView,
+    DriverListView,
+    DriverDetailView,
+    CarDetailView
+)
 
 urlpatterns = [
     path("", index, name="index"),
     path("manufacturers/", ManufacturerListView.as_view(),
          name="manufacturer-list"),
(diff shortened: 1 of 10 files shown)


taxi/urls.py
```diff
@@ -1,9 +1,18 @@
 from django.urls import path
 
-from .views import index
+from .views import (index, CarListView,
+                    ManufacturerListView, car_detail_view,
+                    DriverListView, DriverDetailView)
```
vsmutok:
```suggestion
from .views import (
    index, 
    CarListView,
    ManufacturerListView, 
    car_detail_view,
    DriverListView, 
    DriverDetailView
)
```

taxi/views.py
```diff
@@ -13,3 +15,44 @@ def index(request):
...
+
+def car_detail_view(request: HttpRequest, pk: int) -> HttpResponse:
+    try:
+        car = Car.objects.get(id=pk)
+    except Car.DoesNotExist:
+        raise Http404("Car does not exist")
+    context = {
+        "car": car,
+    }
+    return render(request, "taxi/car_detail.html", context=context)
```
vsmutok: please rewrite in class base view

vsmutok (changes requested):

taxi/views.py
oboliziuk: then I studied and compared the capabilities of the class and the function)

vasylhnatiuk (approved):

PR #788, opened 2024-09-09
implement py-taxi-class-based-views

ValeriiaKyr:
Снимок экрана 2024-09-09 в 16 42 43
Снимок экрана 2024-09-09 в 16 42 57
Снимок экрана 2024-09-09 в 16 43 11
Снимок экрана 2024-09-09 в 16 43 29
Снимок экрана 2024-09-09 в 16 43 40

### DIFF
--- a/taxi/urls.py
+++ b/taxi/urls.py
@@ -19,7 +19,11 @@ urlpatterns = [
     path("cars/", CarListView.as_view(), name="car-list"),
     path("cars/<int:pk>/", CarDetailView.as_view(), name="car-detail"),
     path("drivers/", DriverListView.as_view(), name="driver-list"),
-    path("drivers/<int:pk>/", DriverDetailView.as_view(), name="driver-detail"),
+    path(
+        "drivers/<int:pk>/",
+        DriverDetailView.as_view(),
+        name="driver-detail"
+    ),
 ]
 
 app_name = "taxi"
--- a/taxi/urls.py
+++ b/taxi/urls.py
@@ -17,9 +17,9 @@ urlpatterns = [
         name="manufacturer-list"
     ),
     path("cars/", CarListView.as_view(), name="car-list"),
-    path("cars/<int:pk>", CarDetailView.as_view(), name="car-detail"),
+    path("cars/<int:pk>/", CarDetailView.as_view(), name="car-detail"),
     path("drivers/", DriverListView.as_view(), name="driver-list"),
-    path("drivers/<int:pk>", DriverDetailView.as_view(), name="driver-detail"),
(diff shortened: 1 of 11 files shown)


taxi/urls.py
```diff
@@ -1,9 +1,25 @@
...
     path("", index, name="index"),
+    path(
+        "manufacturers/",
+        ManufacturerListView.as_view(),
+        name="manufacturer-list"
+    ),
+    path("car/", CarListView.as_view(), name="car-list"),
+    path("car/<int:pk>", CarDetailView.as_view(), name="car-detail"),
+    path("drivers/", DriverListView.as_view(), name="driver-list"),
+    path("drivers/<int:pk>", DriverDetailView.as_view(), name="driver-detail"),
```
MNDonut: are you sure these urls are correct?

taxi/views.py
```diff
@@ -13,3 +14,36 @@ def index(request):
     }
 
     return render(request, "taxi/index.html", context=context)
+
+
+class ManufacturerListView(generic.ListView):
+    model = Manufacturer
+    queryset = Manufacturer.objects.all().order_by("name")
```
```
select * from manufacturers
```
```
select * from (
    previous query
)
order by name
```

taxi/views.py
```diff
@@ -13,3 +14,36 @@ def index(request):
...
+    paginate_by = 5
+    template_name = "taxi/manufacturer_list.html"
+    context_object_name = "manufacturer_list"
+
+
+class CarListView(generic.ListView):
+    model = Car
+    paginate_by = 5
+    queryset = (Car.objects.select_related("manufacturer")
```
MNDonut: Correct formatting is
```
queryset = (
    Car.objects.select_related()
    .prefetch_related()
    .order_by()
    .limit()
)
```

MNDonut (changes requested):

taxi/urls.py
```diff
@@ -16,8 +16,8 @@
         ManufacturerListView.as_view(),
         name="manufacturer-list"
     ),
-    path("car/", CarListView.as_view(), name="car-list"),
-    path("car/<int:pk>", CarDetailView.as_view(), name="car-detail"),
+    path("cars/", CarListView.as_view(), name="car-list"),
+    path("cars/<int:pk>", CarDetailView.as_view(), name="car-detail"),
     path("drivers/", DriverListView.as_view(), name="driver-list"),
     path("drivers/<int:pk>", DriverDetailView.as_view(), name="driver-detail"),
```
MNDonut: urls still are incorrect

MNDonut (changes requested):

MNDonut (approved):

PR #790, opened 2024-09-09
first_try

yashkunn:


yashkunn:
image
image
image
image
image
image

taxi/urls.py
```diff
@@ -1,9 +1,17 @@
 from django.urls import path
 
-from .views import index
+from .views import (index, ManufacturerListView, CarListView,
+                    CarDetailView, DriverListView, DriverDetailView)
```
MNDonut: we don't format imports in this way.
1 line - 1 import

taxi/urls.py
```diff
@@ -1,9 +1,17 @@
...
 
 urlpatterns = [
     path("", index, name="index"),
+    path("manufacturers/", ManufacturerListView.as_view(),
+         name="manufacturer-list"),
+    path("cars/", CarListView.as_view(), name="car-list"),
+    path("cars/<int:pk>/", CarDetailView.as_view(), name="car-detail"),
+    path("drivers/", DriverListView.as_view(), name="driver-list"),
+    path("drivers/<int:pk>/", DriverDetailView.as_view(),
+         name="driver-detail"),
```
MNDonut: format these parameters as in usual functions
either all parameters on one line or 1 line - 1 parameter

taxi/views.py
```diff
@@ -13,3 +14,35 @@ def index(request):
...
+    model = Driver
+    template_name = "taxi/driver_list.html"
+    queryset = Driver.objects.all().order_by("last_name", "first_name")
+    paginate_by = 5
+
+
+class DriverDetailView(generic.DetailView):
+    model = Driver
+    template_name = "taxi/driver_detail.html"
+    queryset = Driver.objects.all().prefetch_related("cars")
```
```
select * from manufacturers
```
```
select * from (
    previous query
)
order by name
```

MNDonut (changes requested):

MNDonut (approved):

PR #794, opened 2024-09-09
my page

panichqa:


panichqa:
image
image
image
image
image

taxi/views.py
```diff
@@ -13,3 +14,35 @@ def index(request):
     }
 
     return render(request, "taxi/index.html", context=context)
+
+
+class ManufacturerListView(generic.ListView):
+    model = Manufacturer
+    template_name = "taxi/manufacturer_list.html"
+    queryset = Manufacturer.objects.all().order_by("name")
```
```
select * from manufacturers
```
```
select * from (
    previous query
)
order by name
```

MNDonut (changes requested):

MNDonut (approved):

PR #795, opened 2024-09-10
solutions

HalynaRohatska:
2024-09-10_09h09_55
2024-09-10_09h10_13
2024-09-10_09h10_28
2024-09-10_09h10_39
2024-09-10_09h10_50
2024-09-10_09h11_03

### DIFF
--- a/taxi/urls.py
+++ b/taxi/urls.py
@@ -1,7 +1,7 @@
 from django.urls import path
 
-from .views import (index, ManufacturerListView, CarListView,
-                    CarDetailView, DriverListView, DriverDetailView)
+from .views import index, ManufacturerListView, CarListView
+from .views import CarDetailView, DriverListView, DriverDetailView
 
 urlpatterns = [
     path("", index, name="index"),
--- a/taxi/urls.py
+++ b/taxi/urls.py
@@ -1,7 +1,13 @@
 from django.urls import path
 
-from .views import index, ManufacturerListView, CarListView
-from .views import CarDetailView, DriverListView, DriverDetailView
+from taxi.views import (
+    index,
+    ManufacturerListView,
+    CarListView,
+    DriverListView,
+    CarDetailView,
+    DriverDetailView
+)
 
 urlpatterns = [
     path("", index, name="index"),
--- a/taxi/urls.py
+++ b/taxi/urls.py
@@ -1,10 +1,15 @@
 from django.urls import path
 
-from .views import index, ManufacturerListView, CarListView, CarDetailView, DriverListView, DriverDetailView
+from .views import (index, ManufacturerListView, CarListView,
+                    CarDetailView, DriverListView, DriverDetailView)
 
 urlpatterns = [
     path("", index, name="index"),
-    path("manufacturers/", ManufacturerListView.as_view(), name="manufacturer-list"),
+    path(
+        "manufacturers/",
+        ManufacturerListView.as_view(),
+        name="manufacturer-list"
+    ),
     path("cars/", CarListView.as_view(), name="car-list"),
     path("cars/<int:pk>/", CarDetailView.as_view(), name="car-detail"),
     path("drivers/", DriverListView.as_view(), name="driver-list"),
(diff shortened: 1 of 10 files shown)


taxi/urls.py
```diff
@@ -1,9 +1,19 @@
...
     path("", index, name="index"),
+    path(
+        "manufacturers/",
+        ManufacturerListView.as_view(),
+        name="manufacturer-list"
+    ),
+    path("cars/", CarListView.as_view(), name="car-list"),
+    path("cars/<int:pk>/", CarDetailView.as_view(), name="car-detail"),
+    path("drivers/", DriverListView.as_view(), name="driver-list"),
+    path("drivers/<int:pk>/", DriverDetailView.as_view(), name="driver-detail")
```
MNDonut: we don't format imports in this way. you should format them like the second path above

taxi/views.py
```diff
@@ -13,3 +14,33 @@ def index(request):
     }
 
     return render(request, "taxi/index.html", context=context)
+
+
+class ManufacturerListView(generic.ListView):
+    model = Manufacturer
+    queryset = Manufacturer.objects.all().order_by("name")
+    paginate_by = 5
+
```
```
select * from manufacturers
```
```
select * from (
    previous query
)
order by name
```

taxi/views.py
```diff
@@ -13,3 +14,33 @@ def index(request):
...
+
+class CarListView(generic.ListView):
+    model = Car
+    paginate_by = 5
+    queryset = (Car.objects.
+                select_related("manufacturer").
+                prefetch_related("drivers")
+                )
```
MNDonut: actually move the first line on the next line and the closing bracket should be on the same lvl that "queryset"

taxi/views.py
```diff
@@ -13,3 +14,33 @@ def index(request):
...
+
+
+class CarDetailView(generic.DetailView):
+    model = Car
+
+
+class DriverListView(generic.ListView):
+    model = Driver
+    queryset = Driver.objects.all().order_by("id")
```
MNDonut: django orders models by id automatically(unless to set ordering in Meta class)

MNDonut (changes requested):

taxi/urls.py
```diff
@@ -1,9 +1,19 @@
 from django.urls import path
 
-from .views import index
+from .views import index, ManufacturerListView, CarListView
```
Oleksl888:
```
from taxi.views import (
   index,
   ...

File: taxi/views.py [hunk 6]
```diff
@@ -13,3 +14,34 @@ def index(request):
...
+    paginate_by = 5
+
+
+class CarListView(generic.ListView):
+    model = Car
+    paginate_by = 5
+    queryset = (
+        Car.objects.
+        select_related("manufacturer").
+        prefetch_related("drivers")
```
Oleksl888: are you returning Drivers on a list view? Why prefetch here?

taxi/views.py
```diff
@@ -13,3 +14,34 @@ def index(request):
...
+    )
+
+
+class CarDetailView(generic.DetailView):
+    model = Car
+
+
+class DriverListView(generic.ListView):
+    model = Driver
+    queryset = Driver.objects.all()
```
Oleksl888: why override queryset here without adding any extra logic?

templates/base.html
```diff
@@ -29,6 +29,7 @@
       {% block content %}{% endblock %}
 
       {% block pagination %}
+        {% include "includes/pagination.html" %}
```
Oleksl888: You should add pagination only on pages that have a list of items - no need to use it on detail view pages

Oleksl888 (changes requested):

taxi/views.py
```diff
@@ -27,8 +27,7 @@ class CarListView(generic.ListView):
     paginate_by = 5
     queryset = (
         Car.objects.
-        select_related("manufacturer").
-        prefetch_related("drivers")
+        select_related("manufacturer")
     )
```
MNDonut: You can fit it in 1 line

MNDonut (changes requested):

MNDonut (approved):

PR #796, opened 2024-09-10
Solution

dpiuro:


dpiuro:
image
image
image
image

taxi/urls.py
```diff
@@ -1,9 +1,40 @@
...
-
-from .views import index
+from .views import (
+    ManufacturerListView,
+    CarListView,
+    CarDetailView,
+    DriverListView,
+    DriverDetailView,
+    index
+)
```
MNDonut: split different groups of imports

dpiuro:
image
image
image
image

taxi/views.py
```diff
@@ -1,15 +1,38 @@
 from django.shortcuts import render
-
-from taxi.models import Driver, Car, Manufacturer
+from django.views.generic import ListView, DetailView
+from .models import Car, Driver, Manufacturer
```
MNDonut: fix imports

taxi/views.py
```diff
@@ -1,15 +1,38 @@
...
     }
-
     return render(request, "taxi/index.html", context=context)
+
+
+class ManufacturerListView(ListView):
+    model = Manufacturer
+    queryset = Manufacturer.objects.all().order_by("name")
+    paginate_by = 5
+
```
```
select * from manufacturers
```
```
select * from (
    previous query
)
order by name
```

MNDonut (changes requested):

MNDonut (approved):

PR #797, opened 2024-09-10
task done

Sergi0bbb:
manufacturers list:
image

cars list:
image

drivers list:
image

car details:
image

driver details:
image

### DIFF
--- a/taxi/urls.py
+++ b/taxi/urls.py
@@ -1,6 +1,6 @@
 from django.urls import path
 
-from .views import (
+from taxi.views import (
     index,
     ManufacturerListView,
     CarListView,
--- a/taxi/urls.py
+++ b/taxi/urls.py
@@ -19,7 +19,7 @@ urlpatterns = [
     path("cars/", CarListView.as_view(), name="car-list"),
     path("cars/<int:pk>/", CarDetailView.as_view(), name="car-detail"),
     path("drivers/", DriverListView.as_view(), name="driver-list"),
-    path("drivers/<int:pk>/", DriverDetailView.as_view(), name="driver-detail"),
+    path("drivers/<int:pk>/", DriverDetailView.as_view(), name="driver-detail")
 ]
 
 app_name = "taxi"
--- a/taxi/urls.py
+++ b/taxi/urls.py
@@ -1,9 +1,25 @@
 from django.urls import path
 
-from .views import index
+from .views import (
+    index,
+    ManufacturerListView,
+    CarListView,
+    CarDetailView,
+    DriverListView,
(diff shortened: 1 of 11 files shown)


taxi/urls.py
```diff
@@ -1,9 +1,25 @@
 from django.urls import path
 
-from .views import index
+from .views import (
```
Oleksl888: use full path to module in imports

taxi/views.py
```diff
@@ -13,3 +14,29 @@ def index(request):
     }
 
     return render(request, "taxi/index.html", context=context)
+
+
+class ManufacturerListView(generic.ListView):
+    model = Manufacturer
+    queryset = Manufacturer.objects.all()
```
Oleksl888: why override queryset here without adding any extra logic? (you defined ordering in Meta class so you don't really need to override queryset here)

taxi/views.py
```diff
@@ -13,3 +14,29 @@ def index(request):
...
+
+
+class DriverListView(generic.ListView):
+    model = Driver
+    paginate_by = 5
+
+
+class DriverDetailView(generic.DetailView):
+    model = Driver
+    queryset = Driver.objects.prefetch_related("cars")
```
Oleksl888: >  Optimize query: don't make a query for a **manufacturer** for each car (N+1 problem).
so you need to specify manufacturers field here too

templates/base.html
```diff
@@ -25,15 +24,14 @@
 
     </div>
     <div class="col-sm-10 ">
-
-      {% block content %}{% endblock %}
-
+      {% block content %}
+      {% endblock %}
       {% block pagination %}
+        {% include "includes/pagination.html" %}
```
Oleksl888: would make sense to include pagination only on list views

templates/taxi/car_detail.html
```diff
@@ -0,0 +1,28 @@
+{% extends "base.html" %}
+
+{% block content %}
+  <h1>Car: {{ object.model }}</h1>
```
Oleksl888: why use object here? Here you get `car` object from context and using it would be more obvious and descriptive

templates/taxi/driver_detail.html
```diff
@@ -0,0 +1,22 @@
+{% extends "base.html" %}
+
+{% block content %}
+  <h1>{{ object.username }}</h1>
```
Oleksl888, line 4: `driver`

Oleksl888 (changes requested):

MNDonut (approved):

PR #800, opened 2024-09-10
Solution

AndriiHamasa:


AndriiHamasa:
Снимок экрана 2024-09-10 161242
Снимок экрана 2024-09-10 161300
Снимок экрана 2024-09-10 161441
Снимок экрана 2024-09-10 161458
Снимок экрана 2024-09-10 161521

AndriiHamasa:
Снимок экрана 2024-09-10 161831
Снимок экрана 2024-09-10 161842
Снимок экрана 2024-09-10 161945
Снимок экрана 2024-09-10 162017
Снимок экрана 2024-09-10 162054

taxi/urls.py
```diff
@@ -1,9 +1,25 @@
...
     path("", index, name="index"),
+    path(
+        "manufacturers/",
+        ManufacturerListView.as_view(),
+        name="manufacturer-list"
+    ),
+    path("cars/", CarListView.as_view(), name="car-list"),
+    path("cars/<int:pk>", CarDetailView.as_view(), name="car-detail"),
+    path("drivers/", DriverListView.as_view(), name="driver-list"),
+    path("drivers/<int:pk>", DriverDetailView.as_view(), name="driver-detail"),
```

taxi/views.py
```diff
@@ -13,3 +14,29 @@ def index(request):
...
+
+class ManufacturerListView(generic.ListView):
+    model = Manufacturer
+    queryset = Manufacturer.objects.order_by("name")
+    paginate_by = 5
+
+
+class CarListView(generic.ListView):
+    model = Car
+    queryset = Car.objects.select_related("manufacturer").all()
```
MNDonut: No need to use all. select_related returns queryset

taxi/views.py
```diff
@@ -13,3 +14,29 @@ def index(request):
...
+
+
+class DriverListView(generic.ListView):
+    model = Driver
+    paginate_by = 5
+
+
+class DriverDetailView(generic.DetailView):
+    model = Driver
+    queryset = Driver.objects.prefetch_related("cars").all()
```
MNDonut: no need to use all

MNDonut (changes requested):

taxi/views.py
```diff
@@ -13,3 +14,29 @@ def index(request):
...
+
+
+class DriverListView(generic.ListView):
+    model = Driver
+    paginate_by = 5
+
+
+class DriverDetailView(generic.DetailView):
+    model = Driver
+    queryset = Driver.objects.prefetch_related("cars")
```
Oleksl888: would make sense to prefetch `cars__manufacturer` since it is another table relation on Car

taxi/urls.py
```diff
@@ -1,9 +1,29 @@
 from django.urls import path
 
-from .views import index
+from .views import (
```
Oleksl888: we don't use relative imports in our projects

templates/base.html
```diff
@@ -29,6 +29,7 @@
       {% block content %}{% endblock %}
 
       {% block pagination %}
+        {% include 'includes/pagination.html' %}
```
Oleksl888: pagination should be added only where it is necessary, not all views return a list that needs to be paginated

Oleksl888 (changes requested):

templates/base.html
AndriiHamasa: but in pagination.html file I have if-construction, that check if it is paginated. But ok,
I moved checking to base file in block pagination

templates/base.html
```diff
@@ -6,9 +6,9 @@
   <meta charset="utf-8">
   <meta name="viewport" content="width=device-width, initial-scale=1">
   <link rel="stylesheet"
-      href="https://cdn.jsdelivr.net/npm/bootstrap@4.5.3/dist/css/bootstrap.min.css"
-      integrity="sha384-TX8t27EcRE3e/ihU7zmQxVncDAy5uIKz4rEkgIXeMed4M0jlfIDPvg6uqKI2xXr2"
-      crossorigin="anonymous">
+        href="https://cdn.jsdelivr.net/npm/bootstrap@4.5.3/dist/css/bootstrap.min.css"
```
Oleksl888, line 9: overindented

templates/base.html
```diff
@@ -27,10 +27,11 @@
     <div class="col-sm-10 ">
 
       {% block content %}{% endblock %}
-
       {% block pagination %}
+        {% if is_paginated %}
```
Oleksl888: ` {% if is_paginated %}` this should be inside of `pagination.html`

templates/base.html
```diff
@@ -27,10 +27,11 @@
...
+          {% include 'includes/pagination.html' %}
```
Oleksl888: When you put this include in base.html it means it will be present on all templates that inherit from base.html. It means that it still will be evaluated on every page even if it won't be displayed. Also what is the point of making this an include then if it just a part of base template?
So my suggestion was to include it only on list views - those are the only templates where makes sense to apply pagination

Oleksl888 (changes requested):

templates/base.html
Oleksl888: Still overidented - in html we use 2 spaces for indentation

Oleksl888 (approved):
Well done!

PR #802, opened 2024-09-10
'Solution'

ABaidin:
Screenshot 2024-09-10 165350
Screenshot 2024-09-10 165404
Screenshot 2024-09-10 165416
Screenshot 2024-09-10 165428
Screenshot 2024-09-10 165438
Screenshot 2024-09-10 165458

### DIFF
--- a/taxi/views.py
+++ b/taxi/views.py
@@ -1,9 +1,11 @@
+from django.http import HttpRequest, HttpResponse
 from django.shortcuts import render
+from django.views.generic import ListView, DetailView
 
 from taxi.models import Driver, Car, Manufacturer
 
 
-def index(request):
+def index(request: HttpRequest) -> HttpResponse:
     """View function for the home page of the site."""
 
     context = {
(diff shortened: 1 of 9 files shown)


taxi/views.py
```diff
@@ -13,3 +15,36 @@ def index(request):
...
+
+
+class CarDetailView(DetailView):
+    model = Car
+
+
+class DriverListView(ListView):
+    model = Driver
+    paginate_by = 5
+
```
```
select * from manufacturers
```
```
select * from (
    previous query
)
order by name
```

taxi/views.py
```diff
@@ -13,3 +15,36 @@ def index(request):
...
+
+class DriverDetailView(DetailView):
+    model = Driver
+    queryset = Driver.objects.prefetch_related("cars__manufacturer")
+
+    def get_context_data(self, **kwargs) -> dict:
+        context = super().get_context_data(**kwargs)
+        context["cars"] = self.object.cars.all()
+        return context
```
MNDonut: can't you fetch cars of the driver inside template?

MNDonut (changes requested):

MNDonut (approved):

PR #803, opened 2024-09-10
Solution Politov

Alexandr-Politov:


Alexandr-Politov:
car_detail
car_list
driver_detail
driver_list
manufacturer_list

taxi/urls.py
```diff
@@ -1,9 +1,19 @@
 from django.urls import path
 
-from .views import index
+from .views import (index, ManufacturerListView, CarListView,
+                    CarDetailView, DriverListView, DriverDetailView)
```
MNDonut: fix formatting

taxi/urls.py
```diff
@@ -1,9 +1,19 @@
...
 
 urlpatterns = [
     path("", index, name="index"),
+    path("cars/", CarListView.as_view(), name="car-list"),
+    path("cars/<int:pk>", CarDetailView.as_view(), name="car-detail"),
+    path("drivers/", DriverListView.as_view(), name="driver-list"),
+    path("drivers/<int:pk>", DriverDetailView.as_view(), name="driver-detail"),
```

taxi/views.py
```diff
@@ -1,15 +1,41 @@
...
+
+
+class DriverListView(generic.ListView):
+    model = Driver
+    paginate_by = 5
+
+
+class DriverDetailView(generic.DetailView):
+    model = Driver
+    queryset = Driver.objects.prefetch_related("cars").all()
```
```
select * from manufacturers
```
```
select * from (
    previous query
)
order by name
```

MNDonut (changes requested):

MNDonut (approved):

PR #806, opened 2024-09-10
Solution

Bogdan-Rylov:
Screenshot 2024-09-10 234335
Screenshot 2024-09-10 234429
Screenshot 2024-09-10 234418
Screenshot 2024-09-10 234405
Screenshot 2024-09-10 234347


taxi/models.py
```diff
@@ -1,19 +1,35 @@
...
 
+    class Meta:
+        ordering = ["name"]
+
 
 class Driver(AbstractUser):
     license_number = models.CharField(max_length=255, unique=True)
 
+    class Meta:
+        ordering = ["license_number"]
```
MNDonut, line 18: did you run migrations after these changes?

taxi/urls.py
```diff
@@ -1,9 +1,37 @@
...
+        CarDetailView.as_view(),
+        name="car-detail"
+    ),
+    path(
+        "drivers/", DriverListView.as_view(), name="driver-list"
+    ),
+    path(
+        "drivers/<int:pk>",
+        DriverDetailView.as_view(),
+        name="driver-detail"
```

taxi/views.py
```diff
@@ -1,9 +1,14 @@
+from http.client import HTTPResponse
```
MNDonut: It's a wrong annotation. There is another class in django.http

taxi/views.py
```diff
@@ -13,3 +18,39 @@ def index(request):
...
+    context_object_name = "manufacturer_list"
+    paginate_by = 5
+
+
+class CarListView(generic.ListView):
+    model = Car
+    template_name = "taxi/car_list.html"
+    context_object_name = "car_list"
+    queryset = Car.objects.all().select_related("manufacturer")
+    paginate_by = 5
```
```
select * from manufacturers
```
```
select * from (
    previous query
)
order by name
```

taxi/views.py
```diff
@@ -13,3 +18,39 @@ def index(request):
...
+
+
+class DriverDetailView(generic.DetailView):
+    model = Driver
+    queryset = Driver.objects.prefetch_related(
+        Prefetch(
+            "cars",
+            queryset=Car.objects.select_related("manufacturer")
+        )
+    )
```
MNDonut: Just prefetch_related('field1', 'field2') is enough

MNDonut (changes requested):

taxi/models.py
Oleksl888: No, migrations are not pushed to remote. 
@Bogdan-Rylov when you change default ordering on a table you should run migrations and include them in a commit.

taxi/urls.py
```diff
@@ -1,9 +1,37 @@
 from django.urls import path
 
-from .views import index
+from .views import (
```
Oleksl888: avoid using relative imports

taxi/urls.py
```diff
@@ -1,9 +1,37 @@
...
+    CarDetailView,
+    DriverListView,
+    DriverDetailView,
+)
+
 
 urlpatterns = [
     path("", index, name="index"),
+    path(
+        "manufacturers/", ManufacturerListView.as_view(),
```
Oleksl888: each argument should start from new line

taxi/views.py
```diff
@@ -13,3 +15,34 @@ def index(request):
     }
 
     return render(request, "taxi/index.html", context=context)
+
+
+class ManufacturerListView(generic.ListView):
+    model = Manufacturer
+    template_name = "taxi/manufacturer_list.html"
```
Oleksl888: you are using default template name and context object name in your project so there is no need to set them explicitly here - same with below views

Oleksl888 (changes requested):

Oleksl888 (approved):
Good job! :+1:

PR #808, opened 2024-09-10
Solution

mnezbrytska:


mnezbrytska:
Screenshot 2024-09-11 003111
Screenshot 2024-09-11 003058
Screenshot 2024-09-11 003050
Screenshot 2024-09-10 154434

taxi/urls.py
```diff
@@ -1,9 +1,28 @@
 from django.urls import path
 
-from .views import index
+from .views import (index,
+                    ManufacturerListView,
+                    CarListView,
+                    DriverListView,
+                    CarDetailView,
+                    DriverDetailView)
```
MNDonut: fix formatting

taxi/views.py
```diff
@@ -11,5 +12,29 @@ def index(request):
...
         "num_manufacturers": Manufacturer.objects.count(),
     }
-
     return render(request, "taxi/index.html", context=context)
+
+
+class ManufacturerListView(generic.ListView):
+    model = Manufacturer
+    queryset = Manufacturer.objects.all().order_by("name")
+    paginate_by = 5
```
```
select * from manufacturers
```
```
select * from (
    previous query
)
order by name
```

taxi/views.py
```diff
@@ -11,5 +12,29 @@ def index(request):
...
+    model = Driver
+    paginate_by = 5
+
+
+class CarDetailView(generic.DetailView):
+    model = Car
+
+
+class DriverDetailView(generic.DetailView):
+    model = Driver
```
MNDonut, line 42: image

MNDonut (changes requested):

taxi/urls.py
```diff
@@ -5,7 +5,8 @@
                     CarListView,
                     DriverListView,
                     CarDetailView,
-                    DriverDetailView)
+                    DriverDetailView
+                    )
```
MNDonut: ) should be on the same lvl that "from"

taxi/views.py
```diff
@@ -17,7 +17,7 @@ def index(request):
 
 class ManufacturerListView(generic.ListView):
     model = Manufacturer
-    queryset = Manufacturer.objects.all().order_by("name")
+    queryset = Manufacturer.objects.filter().order_by("name")
```
MNDonut: why filter? you don't need this part at all. read my previous message carefully

MNDonut (changes requested):

mnezbrytska:
Sorry, I just saw "filter() insted all()" and didn't read to end

taxi/urls.py
Oleksl888: not fixed. you have to change like the below:
```
from taxi.views import (
    index,...

File: templates/base.html [hunk 6]
```diff
@@ -29,7 +29,9 @@
       {% block content %}{% endblock %}
 
       {% block pagination %}
+          {% include "includes/pagination.html" %}
       {% endblock %}
```
Oleksl888: include pagination only on templates that need it (list view related)

Oleksl888 (changes requested):

MNDonut (approved):

PR #811, opened 2024-09-11
Solution

Ivan-Shakhman:
...

### DIFF
--- a/taxi/urls.py
+++ b/taxi/urls.py
@@ -1,6 +1,6 @@
 from django.urls import path
 
-from .views import (
+from taxi.views import (
     index,
     ManufacturerListView,
     CarListView,
--- a/taxi/urls.py
+++ b/taxi/urls.py
@@ -1,9 +1,26 @@
 from django.urls import path
 
-from .views import index
+from .views import (
+    index,
+    ManufacturerListView,
+    CarListView,
+    CarDetailView,
+    DriverListView,
+    DriverDetailView
+)
 
 urlpatterns = [
     path("", index, name="index"),
+    path(
+        "manufacturers/",
+        ManufacturerListView.as_view(),
+        name="manufacturer-list"
+    ),
+    path("cars/", CarListView.as_view(), name="car-list"),
+    path("cars/<int:pk>", CarDetailView.as_view(), name="car-detail"),
+    path("drivers/", DriverListView.as_view(), name="driver-list"),
+    path("drivers/<int:pk>", DriverDetailView.as_view(), name="driver-detail"),
+
 ]
 
 app_name = "taxi"
--- a/taxi/urls.py
+++ b/taxi/urls.py
@@ -17,9 +17,13 @@ urlpatterns = [
         name="manufacturer-list"
     ),
     path("cars/", CarListView.as_view(), name="car-list"),
-    path("cars/<int:pk>", CarDetailView.as_view(), name="car-detail"),
+    path("cars/<int:pk>/", CarDetailView.as_view(), name="car-detail"),
     path("drivers/", DriverListView.as_view(), name="driver-list"),
-    path("drivers/<int:pk>", DriverDetailView.as_view(), name="driver-detail"),
+    path(
+        "drivers/<int:pk>/",
+        DriverDetailView.as_view(),
+        name="driver-detail"
+    ),
 
 ]
 
--- a/taxi/views.py
+++ b/taxi/views.py
@@ -41,4 +41,4 @@ class DriverListView(generic.ListView):
 
 class DriverDetailView(generic.DetailView):
     model = Driver
-    queryset = Driver.objects.prefetch_related().order_by("last_name")
+    queryset = Driver.objects.prefetch_related("cars").order_by("last_name")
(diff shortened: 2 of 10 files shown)


Ivan-Shakhman:
image

image

image


image


image

taxi/urls.py
```diff
@@ -1,9 +1,26 @@
 from django.urls import path
 
-from .views import index
+from .views import (
```
Oleksl888: avoid using relative imports

taxi/urls.py
```diff
@@ -1,9 +1,26 @@
...
 
 urlpatterns = [
     path("", index, name="index"),
+    path(
+        "manufacturers/",
+        ManufacturerListView.as_view(),
+        name="manufacturer-list"
+    ),
+    path("cars/", CarListView.as_view(), name="car-list"),
+    path("cars/<int:pk>", CarDetailView.as_view(), name="car-detail"),
```
Oleksl888: add slashes to all url paths

taxi/views.py
```diff
@@ -13,3 +14,32 @@ def index(request):
     }
 
     return render(request, "taxi/index.html", context=context)
+
+
+class ManufacturerListView(generic.ListView):
+    model = Manufacturer
+    queryset = Manufacturer.objects.all().order_by("name")
```
Oleksl888: No need for all() here - just by adding order_by query will be prepared to include all results

taxi/views.py
```diff
@@ -13,3 +14,32 @@ def index(request):
...
+    paginate_by = 5
+    template_name = "taxi/manufacturer_list.html"
```
Oleksl888: you don't need to set template name if you are using default one

taxi/views.py
```diff
@@ -13,3 +14,32 @@ def index(request):
...
+
+
+class CarListView(generic.ListView):
+    model = Car
+    queryset = Car.objects.select_related().order_by("model")
```
Oleksl888, line 27: it is recommended to define fields in select related explicitly. here it is ok because you have only one foreign key but otherwise it may result in unnecessary joins making query more complex:

> There may be some situations where you wish to call select_related() with a lot of related objects, or where you don’t know all of the relations. In these cases it is possible to call select_related() with no arguments. This will follow all non-null foreign keys it can find - nullable foreign keys must be specified. This is not recommended in most cases as it is likely to make the underlying query more complex, and return more data, than is actually needed.

taxi/views.py
```diff
@@ -13,3 +14,32 @@ def index(request):
...
+
+class CarDetailView(generic.DetailView):
+    model = Car
+
+
+class DriverListView(generic.ListView):
+    model = Driver
+    paginate_by = 5
+    template_name = "taxi/driver_list.html"
+    queryset = Driver.objects.select_related().order_by("last_name")
```
Oleksl888: you don't access any related fields on this view so why would you use select related here?

taxi/views.py
```diff
@@ -13,3 +14,32 @@ def index(request):
...
+
+
+class DriverDetailView(generic.DetailView):
+    model = Driver
+    queryset = Driver.objects.select_related().order_by("last_name")
```
Oleksl888: here you need to optimize a call to M2M field `cars` which in turn needs to access manufacturer info on the car. Think what would be more appropriate to use here

templates/base.html
```diff
@@ -29,6 +29,7 @@
       {% block content %}{% endblock %}
 
       {% block pagination %}
+        {% include "includes/pagination.html" %}
```
Oleksl888: include pagination only on tempaltes that show a list of items

templates/taxi/car_detail.html
```diff
@@ -0,0 +1,16 @@
+{% extends "base.html" %}
+
+{% block content %}
+  <h1>Detail info about {{ object.model }}</h1>
```
Oleksl888: why use object instead of model name? wouldn't `car` be more obvious and straightforward?

templates/taxi/car_list.html
```diff
@@ -0,0 +1,16 @@
+{% extends "base.html" %}
+
+{% block content %}
+  <h1>Cars List</h1>
+    <ul>
+      {% if object_list %}
```
Oleksl888: by default it will be `modelname_list` (car_list in this case). Why not use that for readability?

Oleksl888 (changes requested):

taxi/views.py
```diff
@@ -37,9 +36,9 @@ class DriverListView(generic.ListView):
...
     paginate_by = 5
     template_name = "taxi/driver_list.html"
-    queryset = Driver.objects.select_related().order_by("last_name")
+    queryset = Driver.objects.order_by("last_name")
 
 
 class DriverDetailView(generic.DetailView):
     model = Driver
-    queryset = Driver.objects.select_related().order_by("last_name")
+    queryset = Driver.objects.prefetch_related().order_by("last_name")
```
MNDonut: Don't use prefetch_related without args because it may fetch more data than you expect

taxi/urls.py
MNDonut: still not fixed

MNDonut (changes requested):
Some improvements are required

MNDonut (approved):

PR #859, opened 2024-10-31
Solution

vladyslav-tmf:
Home Page:

Screenshot_1

/manufacturers:

Screenshot_3

/cars:

Screenshot_4

/drivers:

Screenshot_5

/cars/{id}:

Screenshot_5

/drivers/{id}:

Screenshot_6

### DIFF
--- a/taxi/views.py
+++ b/taxi/views.py
@@ -1,6 +1,7 @@
 from django.shortcuts import render
+from django.views import generic
 
-from taxi.models import Driver, Car, Manufacturer
+from taxi.models import Car, Driver, Manufacturer
 
 
 def index(request):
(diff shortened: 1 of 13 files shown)


taxi/views.py
```diff
@@ -13,3 +14,29 @@ def index(request):
     }
 
     return render(request, "taxi/index.html", context=context)
+
+
+class ManufacturerListView(generic.ListView):
+    model = Manufacturer
+    queryset = Manufacturer.objects.all().order_by("name")
```
Arsen-hrynevych, line 21: When you use `order_by`, it by default returns all objects that satisfy the rule. Therefore, you don't need to explicitly use `objects.all()`.

templates/taxi/car_list.html
```diff
@@ -0,0 +1,22 @@
...
+    {% if car_list %}
+      <h1 class="center">Cars table:</h1>
+      <tr>
+        <th class="point">Car ID</th>
+        <th class="point">Manufacturer</th>
+        <th class="point">Model</th>
+      </tr>
+      {% for car in car_list %}
+        <tr>
+          <th><a href="{% url 'taxi:car-detail' car.id %}">{{ car.id }}</a></th>
```
Arsen-hrynevych, line 13: The {% url %} template tag generates URLs dynamically using the URL name defined in your urls.py file. The parameters you pass must match the named parameters expected by the URL pattern. 

Example:

templates/taxi/driver_list.html
```diff
@@ -0,0 +1,22 @@
...
+    {% if driver_list %}
+      <h1 class="center">Drivers table:</h1>
+      <tr>
+        <th class="point">Username</th>
+        <th class="point">First name</th>
+        <th class="point">Last name</th>
+      </tr>
+      {% for driver in driver_list %}
+        <tr>
+          <th><a href="{% url 'taxi:driver-detail' driver.id %}">{{ driver.username }}</a></th>
```
Arsen-hrynevych, line 13: The {% url %} template tag generates URLs dynamically using the URL name defined in your urls.py file. The parameters you pass must match the named parameters expected by the URL pattern. 

Example:

Arsen-hrynevych (approved):
Mostly correct. Thanks

PR #864, opened 2024-11-10
solution

dmytrik:
Screenshot 2024-11-10 at 14 25 57
Screenshot 2024-11-10 at 14 26 08
Screenshot 2024-11-10 at 14 26 18
Screenshot 2024-11-10 at 14 26 26
Screenshot 2024-11-10 at 14 26 37

### DIFF
--- a/taxi/views.py
+++ b/taxi/views.py
@@ -1,4 +1,5 @@
 from django.shortcuts import render
+from django.views import generic
 
 from taxi.models import Driver, Car, Manufacturer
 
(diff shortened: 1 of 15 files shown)


taxi/views.py
```diff
@@ -13,3 +14,29 @@ def index(request):
     }
 
     return render(request, "taxi/index.html", context=context)
+
+
+class ManufacturerListView(generic.ListView):
+    model = Manufacturer
+    queryset = Manufacturer.objects.all().order_by("name")
```
Arsen-hrynevych, line 21: When you use `order_by`, it by default returns all objects that satisfy the rule. Therefore, you don't need to explicitly use `objects.all()`.

Arsen-hrynevych (approved):
ok

PR #866, opened 2024-11-12
Develop

KKvladys:
Снимок экрана 2024-11-12 161847
Снимок экрана 2024-11-12 161936
Снимок экрана 2024-11-12 162012
Снимок экрана 2024-11-12 162034
Снимок экрана 2024-11-12 162050


taxi/views.py
```diff
@@ -13,3 +14,28 @@ def index(request):
...
+    model = Car
+
+
+class DriversListView(generic.ListView):
+    model = Driver
+    paginate_by = 5
+
+
+class DriverDetailView(generic.DetailView):
+    model = Driver
```
VladyslavBon, line 41: 5. Create DriverDetailView detail view.
Set model and queryset.
In this view, you display information about cars of the driver. Optimize query: don't make a query for a manufacturer for each car (N+1 problem).

So here you have to add queryset with prefetch_related("cars")

VladyslavBon (approved):

PR #876, opened 2024-11-18
solution

spa7id:


taxi/views.py
```diff
@@ -13,3 +14,36 @@ def index(request):
...
+
+class DriverDetailView(generic.DetailView):
+    model = Driver
+    queryset = Driver.objects.prefetch_related("cars").all()
+
+
+class HomeView:
+    @classmethod
+    def as_view(cls):
+        pass
```
Den-k0: I think it's unnecessary

tests/test_taxi_view.py
Den-k0, line 1: Maybe, you don't need to delete tests 🤔

Den-k0 (approved):

taxi/views.py
spa7id: @Den-k0, u right!
solved

tests/test_taxi_view.py
spa7id: @Den-k0 Thank you, already fixed :)
I have no idea how that happened :)

PR #885, opened 2024-11-19
Solution

MatvijSafonov:
image
image
image
image
image


alauda6 (approved):
I think there should be drivers in your screenshot

PR #896, opened 2024-11-20
solution beta

YaYaYaroslav:


pohgen (changes requested):
fix flake8
fix structure of project (html files need to be in taxi folder except includes files and base.html

PR #990, opened 2025-03-03
Develop

saywin:
Screenshot_1
Screenshot_2
Screenshot_3
Screenshot_4
Screenshot_5


taxi/urls.py
```diff
@@ -1,9 +1,29 @@
 from django.urls import path
 
-from .views import index
+from .views import (
```
SerhiyTo: Use absolute imports instead of relative imports (`from app.models import` instead of `from .models import`)

SerhiyTo (changes requested):

taxi/urls.py
saywin: Done

requirements.txt
```diff
@@ -1,5 +1,7 @@
-django==4.1
+django==4.2.9
 flake8==5.0.4
 flake8-quotes==3.3.1
 flake8-variables-names==0.0.5
 pep8-naming==0.13.2
+
```
SerhiyTo, line 6: remove this empty line

SerhiyTo (approved):

PR #993, opened 2025-03-07
Solution

Baranotik15:


Baranotik15:
image-removebg-preview (3)

taxi/views.py
```diff
@@ -13,3 +14,37 @@ def index(request):
...
+    context_object_name = "manufacturer_list"
+    template_name = "taxi/manufacturer_list.html"
+
+
+class CarListView(ListView):
+    model = Car
+    queryset = Car.objects.select_related("manufacturer").all()
+    paginate_by = 5
+    context_object_name = "car_list"
+    template_name = "taxi/car_list.html"
```
yarzhuravlov: These are default values, no need to set them

taxi/views.py
```diff
@@ -13,3 +14,37 @@ def index(request):
...
+class CarDetailView(DetailView):
+    model = Car
+    template_name = "taxi/car_detail.html"
+
+
+class DriverListView(ListView):
+    model = Driver
+    paginate_by = 5
+    context_object_name = "driver_list"
+    template_name = "taxi/driver_list.html"
```
yarzhuravlov: These are default values, no need to set them

taxi/views.py
```diff
@@ -13,3 +14,37 @@ def index(request):
...
+
+
+class DriverDetailView(DetailView):
+    model = Driver
+    queryset = Driver.objects.prefetch_related("cars__manufacturer").all()
+    template_name = "taxi/driver_detail.html"
```
yarzhuravlov: This is default value, no need to set it

yarzhuravlov (changes requested):
Good job, but some of the code can be omitted

PR #1020, opened 2025-03-30
Implement pages for models

PrimeGlorious:
image
image
image
image
image


Yurkiv313 (approved):
gj

taxi/urls.py
```diff
@@ -1,9 +1,41 @@
...
 
-from .views import index
+from .views import (
+    index,
+    ManufacturerListView,
+    CarListView,
+    CarDetailView,
+    DriverListView,
+    DriverDetailView
+)
```
Boyok1337, line 10: It`s better to use absolut imports instead of relative

Boyok1337 (approved):
Approve, but fix imports

PR #1024, opened 2025-04-01
Anton Zhukov - Solution

windings-lab:


windings-lab:
@mateacademy-ai-mentor 
Driver does not have any fields that would create N+1 problem. So it's okay

PR #1032, opened 2025-04-04
Solution

HappyDen08:
photo_5440743276857650362_w
photo_5440743276857650361_w
photo_5440743276857650360_w


templates/taxi/driver_detail.html
OlenVe, line 1: You need create driver detail template:
            place information about all cars of the driver.

OlenVe (changes requested):
Code and site looks good but you missed  "driver detail template"

PR #1039, opened 2025-04-04
solution

cemen2009:
image_2025-04-05_00-11-34
image_2025-04-05_00-12-00
image_2025-04-05_00-12-24
image_2025-04-05_00-12-47
image_2025-04-05_00-13-09


OlenVe (approved):
Perfect !!

PR #1046, opened 2025-04-06
solution

Anollinna:


Roksolana-K (approved):
Nice job!
But wasn't it necessary to make car-list and driver-list as a table?

PR #1048, opened 2025-04-07
Solution py-taxi-service-class-based-views

Roksolana-K:
image
image
image
image
image
image


taxi/urls.py
```diff
@@ -1,9 +1,37 @@
 from django.urls import path
 
-from .views import index
+from .views import (
```
Boyok1337, line 3: Better to use absolute imports

taxi/views.py
```diff
@@ -13,3 +14,29 @@ def index(request):
     }
 
     return render(request, "taxi/index.html", context=context)
+
+
+class ManufacturerListView(generic.ListView):
+    model = Manufacturer
+    queryset = Manufacturer.objects.all().order_by("name")
```
Boyok1337, line 21: Its better to overwrite get_queryset method.
And you set ordering for Manufacturer model, so u don`t need to double check it here

Boyok1337 (approved):

taxi/urls.py
Roksolana-K: Hello! When I forked the project the import was already there, written in this way. 
But you are right! It's my fault that I didn't double check everything. 
I will try to be more attentive next time 👍

PR #1049, opened 2025-04-08
solution

ntkachenko21:


ntkachenko21:
image

taxi/urls.py
```diff
@@ -1,9 +1,45 @@
 from django.urls import path
 
-from .views import index
+from .views import (
```
Boyok1337, line 3: Better to use absolute imports

taxi/views.py
```diff
@@ -13,3 +14,29 @@ def index(request):
     }
 
     return render(request, "taxi/index.html", context=context)
+
+
+class ManufacturerListView(ListView):
+    model = Manufacturer
+    queryset = Manufacturer.objects.all().order_by("name")
```
Boyok1337, line 21: It`s better to overwrite get_queryset func to use in future

Boyok1337 (approved):
Fix comments

PR #1081, opened 2025-05-14
implemented "py-taxi-service-class-based-views"

Viktor-Stadnik:


Viktor-Stadnik:
SC1
SC2
SC3
SC4
SC5
SC6

PR #1150, opened 2025-08-11
Solution

VadymKyrylko:
image
image
image
image


Orixara (approved):
It's strange that you don't have any cars, manufacturers etc. Did you load data from json?

PR #1174, opened 2025-08-12
Solution

bohdankolisnyk04:


taxi/views.py
```diff
@@ -13,3 +14,31 @@ def index(request):
...
+
+class ManufacturerListView(generic.ListView):
+    model = Manufacturer
+    queryset = Manufacturer.objects.all().order_by("name")
+    paginate_by = 5
+
+
+class CarListView(generic.ListView):
+    model = Car
+    queryset = Car.objects.select_related("manufacturer").all()
```
SensibleN00B, line 27: in this case "all()" is unnecessary

taxi/views.py
```diff
@@ -13,3 +14,31 @@ def index(request):
...
+    paginate_by = 5
+
+
+class CarDetailView(generic.DetailView):
+    model = Car
+
+
+class DriverDetailView(generic.DetailView):
+    model = Driver
+    queryset = Driver.objects.prefetch_related("cars").all()
```
SensibleN00B, line 37: in this case "all()" is unnecessary

templates/base.html
```diff
@@ -28,8 +28,6 @@
 
       {% block content %}{% endblock %}
 
-      {% block pagination %}
-      {% endblock %}
```
SensibleN00B, line 32: it would be better to leave the pagination in the "base.html"

SensibleN00B (approved):
not bad, but you can do this better

PR #1198, opened 2025-09-14
implemented py-taxi-service-class-based-views v1

mishagitcode:


mishagitcode:
image

image

image

image

image

image

mishagitcode:
The Car model defines drivers = models.ManyToManyField(Driver, related_name="cars"), so the correct forward accessor is object.drivers.all. This avoids driver_set and makes the relationship more readable.

PR #1244, opened 2025-11-11
Taxi service class-based views

mtotsevych:


mtotsevych:
Regarding the use of username in the driver list: I initially implemented the detail view URL using username as the identifier, as specified in the task. However, the provided test suite expects the detail view to resolve using the driver's id (primary key), not username. When using username, the tests fail due to a mismatch in the URL pattern and the arguments passed in reverse().

To ensure all tests pass successfully, I reverted the URL pattern and template logic to use id instead of username.

PR #1255, opened 2025-12-03
'Solution'

Vorongor:
Manufacturer list:
mf_list
Car list:
car_list
Car detail:
car_detail
Driver list:
driver_list
Driver detail:
driver_detail


n-korobko (approved):
Good job!!!!)))

PR #1260, opened 2025-12-08
Solution

dmytro-malyk-dm:
Home:
chrome_73FW0rl84C
Manufacturer:
chrome_yKlweZA2us
Cars:
chrome_6Pr0w7ufvS
Drivers:
chrome_jVET95AE4R
Car detail:
chrome_YV8olDo7RG
Driver detail:
chrome_sbZkAMLKos


templates/taxi/driver_detail.html
```diff
@@ -0,0 +1,25 @@
...
+
+{% block content %}
+  <h1>{{ driver.username }}</h1>
+  <ul>
+    <li>First name: {{ driver.first_name }}</li>
+    <li>Last name: {{ driver.last_name }}</li>
+    <li>License Number: {{ driver.license_number }}</li>
+  </ul>
+  <h2>Cars</h2>
+  {% if driver.cars.all %}
```
leemonch1k-space, line 11: In this case, "if" always returns True, you can replace it with for...empty

templates/taxi/car_detail.html
```diff
@@ -0,0 +1,24 @@
+{% extends "base.html" %}
+{% block content %}
+  <h1>{{ car.model }}</h1>
+  <h2>Manufacturer Information</h2>
+  <ul>
+    <li>Name: {{ car.manufacturer.name }}</li>
+    <li>Country: {{ car.manufacturer.country }}</li>
+   </ul>
+  <h2>Drivers</h2>
+  {% if car.drivers.all %}
```
leemonch1k-space, line 10: In this case, "if" always returns True, you can replace it with for...empty

leemonch1k-space (approved):
Good job!

PR #1278, opened 2025-12-11
Taxi service class-based views

ohorodnykostap:
зображення


Akcolotle (approved):
approve, but i don't see sidebar.html with links to the home page, manufacturer list page, car list page, drivers list page.

PR #1282, opened 2025-12-12
Checking tests

oleksii-shcherbak-code:


taxi/views.py
```diff
@@ -13,3 +15,29 @@ def index(request):
     }
 
     return render(request, "taxi/index.html", context=context)
+
+
+class ManufacturerListView(generic.ListView):
+    model = Manufacturer
+    queryset = Manufacturer.objects.all().order_by("name")
+    page_size = 5
```
tigersaymon: wrong variable, should be paginate_by

tigersaymon (changes requested):
Flake8 failed and I can't see any templates or files except views.py.

PR #1298, opened 2025-12-23
Py-taxi-service-class-based-views

Ann-Riabykina:


Ann-Riabykina:
image

taxi/views.py
```diff
@@ -13,3 +13,31 @@ def index(request):
     }
 
     return render(request, "taxi/index.html", context=context)
+
+
+class ManufacturerListView(ListView):
+    model = Manufacturer
+    queryset = Manufacturer.objects.all().order_by("name")
```
oliasymak24: this code will work the same without .all()

oliasymak24 (changes requested):

oliasymak24 (approved):

PR #1366, opened 2026-04-11
Implement django taxi

adonaisantos694:


adonaisantos694:
Car Detail
Cars List
Driver Detail
Drivers
Manafacturers

PR #1469, opened 2026-07-24
fix3

ivansinickij29-cpu:
image

### DIFF
--- a/taxi/urls.py
+++ b/taxi/urls.py
@@ -1,9 +1,30 @@
 from django.urls import path
 
-from .views import index
+from .views import (
+    index,
+    ManufacturerListView,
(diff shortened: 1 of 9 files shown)


taxi/urls.py
```diff
@@ -1,9 +1,30 @@
 from django.urls import path
 
-from .views import index
+from .views import (
```
hladiivolodymyr, line 3: use absolute imports instead of relative imports

hladiivolodymyr (changes requested):
Please fix the flake8 issues and make sure all checks pass locally before pushing your changes.

vsmutok (approved):
Great!

PR #1481, opened 2026-08-03
Solution

vasyl-main-dev:


vasyl-main-dev:
Also i created manufacturer-detail page as bonus

image
image
image
image
image
image
image

taxi/urls.py
```diff
@@ -1,9 +1,20 @@
 from django.urls import path
 
-from .views import index
+from .views import (index, ManufacturerListView, CarListView,
+                    CarDetailView, DriverListView, DriverDetailView,
+                    ManufacturerDetailView)
```
hladiivolodymyr: Please use absolute imports instead of relative imports. For example, use from taxi.views import ... instead of from .views import .... 
Additionally, consider formatting the import for better readability by placing each imported class on a separate line.

hladiivolodymyr (changes requested):

Sn1kls (approved):
Approve – nice work!

